PR #2, opened 2015-07-31
Python3

justcalamari:
Ran 2to3 on all code in the repo. Manually checked all of the code and fixed any remaining errors that I found. Ran diffpy.utils.tests.run and it worked. Could not get wx to work, however.


pavoljuhas:
Cleaned up commits from the backup files.  Adjusted commit messages to git conventions.
Applied as e10bb9dce873b9c5944e8d449b1127a79065c10a.

PR #6, opened 2017-03-24
PY3 - update package metadata

pavoljuhas:
ff merge if it works.


pavoljuhas:
Can we avoid merge bubbles for simple updates that can be fast-forwarded?

```sh
$ git log --graph --oneline 982ebaf~2..982ebaf
*   982ebaf Merge pull request #6 from pavoljuhas/py3-update-package-metadata
|\  
| * 00c6cd6 Remove travis test for Python 3.3.
| * d63bbc2 List Python 3 versions that are supported.
| * de7433e Avoid travis failure due to a bug in conda-build.
|/  
* 65a9bd9 Adjust for conda-build 2.1.4.
```

pavoljuhas:
> What do you propose as an alternative?

I prefer to merge PR-s as fast-forwards when possible.  Such approach keeps history (and log output) linear, simple and without clutter of unnecessary merge commits.  The green merge button on PR page is convenient, but makes it impossible to check the resulting merge before publishing, produces merge bubbles instead of fast-forwards and may give useless author field if user did not setup his email and name on GH correctly.  I don't use the GH merge button at all, it is basically a limited dumbed down functionality of a stock git.

Here is how some PR can be integrated as fast forward:

```sh
# origin is the diffpy/diffpy.utils repository on GH
# local branch python3 tracks origin/python3
git checkout python3
# make sure python3 is in sync with the upstream origin/python3
git pull origin --ff-only
# merge #6.  Per git defaults it would be a fast-forward when descendant of python3.
git pull origin pull/6/head
# have one last look at the new commits to be published ...
# and finally publish:
git push origin python3
```

PR #13, opened 2023-07-19
Load headers in loadData

### LINKED ISSUE #12: pkg_resources deprecated
version.py affected; causes pytest warnings in dependent modules

Sparks29032:
First commit attempts to remove a deprecation (fixes #12). If does not work, will revert.


sbillinge (commented):
This PR is doing 2 things....the header parsing and the pkg_resources.   Let's close this and put those on separate PRs.

In the mean-time, I want to take this opportunity to change the default branch from master to main which means you will have to run a couple of commands to reconnect your local.

It is a good time to do it before we start opening PRs and working on here.

PR #20, opened 2023-07-25
Add header metadata parsing to loaddata

Sparks29032:
Can be merged independently of deprecation fixes in PR #19.


src/diffpy/utils/parsers/loaddata.py
```diff
@@ -26,6 +26,9 @@ def loadData(filename, minrows=10, **kwargs):
     filename -- name of the file we want to load data from.
     minrows  -- minimum number of rows in the first data block.
                 All rows must have the same number of floating point values.
+    headers  -- return also a dictionary of parameters specified in header
```
sbillinge, line 29: can we have a bit more.  For example, this should be a bool and the behavior explained if it is set to False or True nad that the default is False.

Please revisit the other help statements (even the ones written by others) to try and make the clearer like that if htey need it.

sbillinge (commented):
This looks good.

I am a tad worried about a performance overhead if it is doing this check on every line that contains data (there are many such lines).  Please can you time it on some files with header set to false and to true (and against the code unmodified)?

If the performance is not affected we are ok, if it is, then I have some ideas to speed it up.  You may too.

src/diffpy/utils/parsers/loaddata.py
```diff
@@ -98,6 +134,11 @@ def countcolumnsvalues(line):
             # in case of trailing delimiters.
             kwargs.setdefault('usecols', list(range(ncvblock[0])))
             rv = loadtxt(fid, **kwargs)
+
+    # return headers if requested
+    if headers:
+        return hdata, rv
+    # otherwise do not
```
sbillinge, line 141: do we need an else: here?

sbillinge (commented):
let's ensure backwards compatibility.  This function is widely use din the community and we can accidentally break a lot of things if we change the signature.

One thing to do is to roll the new variables into the **kwargs to keep the arglist the same.  Having an if else on the return kind of handles backwards compatibility on the output side but is a bit yucky.

We could also think about a `load_data_and_headers()` function.  It might be safer.  I think that Pavol's design philosophy was that many datafiles containing data have some header text then multiple columns of numbers and so his function has really great utility because it is not for reading gr files per se, but any file that has multi-column data but other stuff hanging around in there of whatever format.

We are kind of mixing the philosophy here, with code that is rather specific to .gr files (some other may have similar formats in the header but others not).

Thinking further it might be nice to have a `load_header()` function and then we could run 
```
data = loadData(file)
metadata = load_header(file)
```
there is a certain clarity to that.   We could eithe rmake a bunch of them, e.g., `load_header_diffpy()`, `load_header_<sthg_else>()` or we could use something like a factory pattern

Sparks29032:
> This looks good.
> 
> I am a tad worried about a performance overhead if it is doing this check on every line that contains data (there are many such lines). Please can you time it on some files with header set to false and to true (and against the code unmodified)?
> 
> If the performance is not affected we are ok, if it is, then I have some ideas to speed it up. You may too.

Yes, I can do some timing, but I think you are right there should be a separate function for header and loading data

Sparks29032:
> let's ensure backwards compatibility. This function is widely use din the community and we can accidentally break a lot of things if we change the signature.
> 
> One thing to do is to roll the new variables into the **kwargs to keep the arglist the same. Having an if else on the return kind of handles backwards compatibility on the output side but is a bit yucky.
> 
> We could also think about a `load_data_and_headers()` function. It might be safer. I think that Pavol's design philosophy was that many datafiles containing data have some header text then multiple columns of numbers and so his function has really great utility because it is not for reading gr files per se, but any file that has multi-column data but other stuff hanging around in there of whatever format.
> 
> We are kind of mixing the philosophy here, with code that is rather specific to .gr files (some other may have similar formats in the header but others not).
> 
> Thinking further it might be nice to have a `load_header()` function and then we could run
> 
> ```
> data = loadData(file)
> metadata = load_header(file)
> ```
> 
> there is a certain clarity to that. We could eithe rmake a bunch of them, e.g., `load_header_diffpy()`, `load_header_<sthg_else>()` or we could use something like a factory pattern

Yep, I did a bit of back compatibility checking using pdfmorph. I cannot put in **kwargs since those are fed into a numpy load data function. I also made sure all new parameters were optional.

src/diffpy/utils/parsers/loaddata.py
Sparks29032: no need for else statement since the return logic takes care of it

sbillinge: OK.  It may still be good to have it just so the intent is clear to future developers like me who can't quite remember the default return logic... ;)  I guess it returns None in that case, right?

sbillinge: my guiding philosophy on htings like that is to always do things to save human brain time, at the expense of extra work for the computer, unless there is a significant performance hit.....   As you do more programming you realize the human brain time is way more valuable and limited resource....

Sparks29032:
> let me know when this is ready to go in

Oops, forgot to unmark

src/diffpy/utils/parsers/loaddata.py
```diff
@@ -16,29 +16,40 @@
...
-    kwargs   -- keyword arguments that are passed to numpy.loadtxt
-
-    Return a numpy array of the data.
-    See also numpy.loadtxt for more details.
+    filename    -- name of the file we want to load data from.
+    minrows     -- minimum number of rows in the first data block.
+                   All rows must have the same number of floating point values.
+    headers     -- when true (default False), the function instead returns
+                   a dictionary of values specified in header.
+    hdel        -- (only used when headers enabled) delimiter for parsing header
```
sbillinge, line 31: please give examples to make this clearer

src/diffpy/utils/parsers/loaddata.py
```diff
@@ -16,29 +16,40 @@
...
+                   information.
+    hignore     -- (only used when headers enabled) ignore header rows beginning
```
sbillinge, line 33: This could also use a short example.

src/diffpy/utils/parsers/loaddata.py
```diff
@@ -16,29 +16,40 @@
...
+                   with any elements in the hignore list
+    kwargs      -- keyword arguments that are passed to numpy.loadtxt including
+                   the following arguments below.
+    delimiter   -- delimiter for the data in the block.
```
sbillinge, line 37: does this have a default?

src/diffpy/utils/parsers/loaddata.py
```diff
@@ -16,29 +16,40 @@
...
+    usecols     -- zero-based index of columns to be loaded, by default use
```
sbillinge, line 38: an example could help here

sbillinge (commented):
looks good.  Small comments on the docstrings before we merge.

Please also edit a changelog somewhere (or make a note where a changelog will be updated)
also, make an issue to update the documentation with this new feature (then work on it)

src/diffpy/utils/parsers/loaddata.py
Sparks29032: Oops, yes I will include the defaults

sbillinge:
Wonderful!,  This is great!  One thing possibly missing from the Changelog is that this is compatible with 3.11 (is it tested)?  But that can be fixed in any PRs during teh release process.

PR #24, opened 2023-07-27
Allow convert of PDF to json

Sparks29032:
Branched from loaddata_headers, which is where the first two commits are from.

For use in the REST API. Before merging, need to address PR #20 issues. However, each function in `loadmetafile.py` depends only on `hddata` and `rv`, so implementation of those functions will not be changed by changes to `loadData()`. However, tests will have to change.

Currently, only supported markup language in `loadmetafile.py` is json.

`load_PDF_into_db()` adds data/metadata from a PDF into a json database

`markup_PDF()` puts a PDF in json format for the REST API

`apply_schema()` reformats a json file to conform to a sample json file


sbillinge:
I will wait to review this until after the load_headers and loadData are separated.

Sparks29032:
Not sure why the differences are not compared with the current main and loadmetadata, but with the old main.

CHANGELOG.md
```diff
@@ -1,20 +1,32 @@
 # Release notes
```
Sparks29032, line 1: Same as loaddata_headers

src/diffpy/utils/parsers/loaddata.py
```diff
@@ -16,29 +16,47 @@
 import numpy
```
Sparks29032, line 16: Same as loaddata_headers

src/diffpy/utils/tests/test_loaddata.py
```diff
@@ -9,6 +9,7 @@
 from diffpy.utils.tests.testhelpers import datafile
```
Sparks29032, line 9: Same as loaddata_headers

src/diffpy/utils/tests/testdata/loaddatawithheaders.txt
Sparks29032, line 1: Same as loaddata_headers

CHANGELOG.md
```diff
@@ -1,20 +1,32 @@
...
 
-## Version 3.1.0 – 2022-12-09
+## Version 3.2.0 – 2023-8-**
```
sbillinge, line 3: This probably needs to be in ISO format.....2023-08-**

CHANGELOG.md
```diff
@@ -1,20 +1,32 @@
...
 ### Added
 
-- Compatibility with Python 3.10, 3.9, 3.8
+- CI Coverage.
+- New tests for loadData function.
 
 ### Changed
 
-### Deprecated
+- loadData function now toggleable. Can return either (a) data read from data blocks or (b) header
```
sbillinge, line 12: This is an add not a deprecation.  YOu haven't decprecated anything (i.e., everything that worked before still works and will continue to work)

CHANGELOG.md
```diff
@@ -1,20 +1,32 @@
...
+information stored above the data block.
+- Exclude wx from tests.
```
sbillinge, line 14: The changelog is mostly for users so we don't need this wx thing here either.

src/diffpy/utils/parsers/loaddata.py
```diff
@@ -16,29 +16,47 @@
...
-                unpacking such as  x, y = loadData(FILENAME, unpack=True).
-                Note transposing the loaded array as loadData(FILENAME).T
-                has the same effect.
-    kwargs   -- keyword arguments that are passed to numpy.loadtxt
-
-    Return a numpy array of the data.
-    See also numpy.loadtxt for more details.
+    The data block is identified as the first matrix block of at least minrows rows
+    and constant number of columns. This seems to work for most of the datafiles including
+    those generated by PDFGetX2.
```
sbillinge, line 24: since we are changing.... say here "....datafiles generated by diffpy programs.

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,212 @@
...
+##############################################################################
+
+import pathlib
+import json
+
+# FIXME: add support for yaml, xml
+supported_formats = ['.json']
+
+
+def load_PDF_into_db(dbname, pdfname, hddata: dict, rv: list, show_path=True):
```
sbillinge, line 23: this name is a bit confusing.  "load" using means "import into memory", like `json.load`.   Can you think carefully what this function does and give it a name that captures that clearly?  Also, update the docstring accordingly.

you talk about "database file" but such a thing doesn't exist.  Do you mean a json format file?

We often write clients that use CRUD commands and the client can be changed depending on the backend.  If the backend is a file-syste, we select a file-system client and write a json file.  IF the backend is a mongoDB database we select the mongodb client and then write code that connects to the database and inserts the information and so on.  That may be too complicated for what we want to do here, but make the docstrings clear so that a future programmer writing a more fancy client manager can immediately figure out what is happening and how to fix it up.  i.e., don't mention "database" anywhere in this function.

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,212 @@
...
+    """Load an entry consisting of PDF header and base data into a database file.
+
+    Requires hdata and rv generated from loadData.
+
+    dbname          -- name of the database file to load an entry into.
+    pdfname         -- name of the PDF file.
```
sbillinge, line 29: why are you limiting it to a pdf file?  couldn't it be any file that has the right header/column data structure?  iq file? chi file? sq file? etc.  Make it as general as possible so it is more reusable.  Basically, any file that loadData can read....

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,212 @@
...
+    hddata  -- Dictionary of metadata.
+    rv      -- List of (r, gr) pairs.
+    muname  -- file to save into (default None, no saving occurs).
+
+    Returns the dictionary loaded from/into markup file.
+    """
+
+    # gather data
+    data = {}
+    data.update({'r': list(rv[:, 0]), 'gr': list(rv[:, 1])})
```
sbillinge, line 117: since we are making it more abstract than just pdf files, maybe change to x and y instead of r and g?

Also, what happens if loadData reads more columns than just 2.  Do we lose the other columns?

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,212 @@
...
+    data.update(hddata)
+
+    # return directly
+    if muname is None:
+        return data
+
+    # save to disk when enabled
+    extension = pathlib.Path(muname).suffix
+    if extension not in supported_formats:
+        raise Exception(f"Format of {muname} is not supported.")
```
sbillinge, line 127: we could also just let loadData decide what it can't read and it can raise an exception? Otherwise this can work?

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,212 @@
...
+                pdfs = json.load(json_read)
+                pdfs.update(entry)
+            with open(dbname, 'w') as json_write:
+                # dump to string first for formatting
+                json.dump(pdfs, json_write, indent=2)
+
+    return pdfs
+
+
+def markup_PDF(hddata: dict, rv: list, muname=None):
```
sbillinge, line 104: this is probably good here in diffpy.utils as long as it is sufficiently abstract as discussed in my comments.

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,212 @@
...
+                    s_flag = False
+                    updated_line = f"{updated_line[:-1]}{line.strip()}\n"
+                    json_file.write(updated_line)
+                elif s_flag:
+                    updated_line += f"{line[:-1].strip()} "
+                else:
+                    json_file.write(line)
+
+
+def apply_schema_to_file(filename, schemaname, multiple_entries=False):
```
sbillinge, line 172: this one may be better in the api where we will maintain the schema

sbillinge (commented):
please see comments

CHANGELOG.md
Sparks29032: It is under changed, not deprecation

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: Got it, agreed

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: Exception raised when trying to put our dictionary in a markup language that we do not have write support for. Currently, I have only put json support for the REST API, but we can later extend to yaml or xml.

Separate from loadData which would deal with read rather than write support.

CHANGELOG.md
sbillinge: ah, sorry, my bad.  Yes, I would put it as an Add (new functionality, old functionality not changed)

CHANGELOG.md
```diff
@@ -1,20 +1,32 @@
...
 
 ### Removed
 
-- Remove the support for Python 3.5, 3.6.
+- Remove use of pkg_resources (deprecated).
+- No longer use Travis.
 
-### Fixed
+## Version 3.1.0 – 2022-12-09
```
sbillinge, line 21: Thanks for doing this :)

src/diffpy/utils/parsers/loaddata.py
sbillinge: I am not sure what this means, but I think it means you are just moving code over from the old function?  Anyway, no need to respond unless you need a response from me and I am not providing it.

src/diffpy/utils/parsers/loadmetafile.py
sbillinge: sorry, I misunderstood.   A clearer docstring maybe would have helped there.

When we move things from memory to more permanent storage we call it serialization.   It is better to use this language.  I am not sure if json is strictly a markup language, it is more like a set of standards for serializing code, so I might stay away from that language too.   The function seems to be doing two things.  One is taking data and putting into a dictionary format and passing it back.  The other is that it is serializing it to disc.  I recommend to split these two functionalities into two separate functions.  One creates the data object (in this case a dict) and the other does serialization, in this case to json.   It is clearer if they are in different functions.

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,233 @@
...
+    # dumps into file, automatically overwrites
+    if extension == '.json':
+        with open(muname, 'w') as json_write:
+            json.dump(data, json_write, indent=2)
+
+    return data
+
+
+def markup_oneline(filename):
+    """Reformat lists in markup languages to take up only one line.
```
sbillinge, line 138: I am not sure we need this function.  Making json readable by humans may not be worth it.  But it is ok to have it here I guess.

sbillinge (commented):
few comments on your comments.

This is turning into a big PR.  In general, try and keep it to one function per PR.....

Hopefully we can get it merged....

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: Got it, made it serialize instead since it refers specifically to serial languages like json, xml, and yaml.

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: No longer has limit, but has an extra dt_colnames input requirement since before we were assuming column names as 'r' and 'gr'.

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: Yes, changed. See dt_colnames parameter in docstring.

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: Thanks for informing that it is called serialization. Changed the nomenclature. Was a bit concerned about what term to use. Very helpful o.O :)

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: I removed since we can perform the same function with our new serialize_data function.

src/diffpy/utils/parsers/loadmetafile.py
Sparks29032: Yea, probably not worth, can remove.

Sparks29032:
Made docstring much more verbose and overhauled `loadmetafile.py`.

Did change naming in loadData since I did not like the return just being called rv (standing for simply return value). Changed to data_table instead. Changed usages of rv from loadData to data_table as well.

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,242 @@
...
+##############################################################################
+
+import pathlib
+import json
+
+# FIXME: add support for yaml, xml
+supported_formats = ['.json']
+
+
+def serialize_data(filename, hdata: dict, data_table: list, show_path=True, dt_colnames=None, serial_file=None):
```
Sparks29032, line 23: New changes start here. No longer called db, now serialize.

src/diffpy/utils/parsers/loadmetafile.py
```diff
@@ -0,0 +1,242 @@
...
+    extension = pathlib.Path(filename).suffix
+    if extension not in supported_formats:
+        raise Exception(f"Format of {filename} is not supported.")
+
+    # json
+    if extension == '.json':
+        with open(filename, 'r') as json_file:
+            j_dict = json.load(json_file)
+
+    return j_dict
```
Sparks29032, line 143: New changes end here. All other changes are just refactoring or removal.

sbillinge (commented):
This looks good.  I guess left to do is 

- [ ] remove  serial_oneline
- [ ] move def `apply_schema_to_file` to api repo
- [ ] see if any of the lines of code not covered by tests can be easily tested, for example, with new mark.parameterize examples.

Then I can merge.

Sparks29032:
All three tasks should be completed. Changed `loadmetadata.py` to `serialization.py`.

Sparks29032:
Not sure why codecov not updating but we have 100% coverage in serialization (see image)
Screenshot_20230811_101456_Chrome

sbillinge:
@Sparks29032 great job, this is really nice code.

My comment from before was I think slightly misunderstood.   In your serialize and deserialize functions there are two things being done really, as I understand it.  On is validating data (checking the shape, the number and length of columns and mapping to values describing what they contain) and the other is the actual serialization (dumping to file in some format that is easy to reload).     What I think is better is if we separate these functionalities into two different functions because I may at some point want to validate some data that has been handed to me but I don't want to dump it to disc.  The best way to do that would be:
```
validate(data)
```
if I don't want to dump, and
```
serialize(data) 
```
if I do want to dump.   The serialize function would be the same as your current function but would look more like
```
def serialize(data):
    validate(data)
    all the code to do the dumping
```
then all the code you have for validation is in `validate()`
if that makes sense.  I merged because the code here is all correct and nice and it should be merged, but I made an issue to do this separation.  We can do it now or just wait for later (maybe wait for later).

PR #25, opened 2023-07-28
Strumining processing

Sparks29032:
For @jaewon078, do not merge.


src/diffpy/utils/tests/test_strumining_helpers.py
```diff
@@ -0,0 +1,42 @@
...
+    hddata = strumining_processing(hddata)
+    markup_PDF(xray_generated, hddata, rv)
+    apply_schema(xray_generated, sm_schema)
+
+    # check against target files
+    assert filecmp.cmp(neutron_generated, neutron_target)
+    assert filecmp.cmp(xray_generated, xray_target)
+
+    # cleanup
+    os.remove(neutron_generated)
```
sbillinge, line 41: this is not a great way to handle things in tests.   Use the tmpfile pytest fixture.    This creates a temporary directory and you can write things there, then read them and do hte asserts etc.   This is then created and removed cleanly even if hte tests fail or error (at least if I understand correclty what you are doing here).

sbillinge (commented):
I left small comments.  Looking good.  Are you moving things out of loadData in the end?  

Just let me know what you want to do with these PRs

src/diffpy/utils/tests/test_strumining_helpers.py
Sparks29032: Got it, that's what I was going for

Sparks29032:
> I left small comments. Looking good. Are you moving things out of loadData in the end?
> 
> Just let me know what you want to do with these PRs

Yep, I am making a `load_header_data` class. I have a current rudimentary function sharing a lot from `loadData` that I tested against these functions, which seems to work, but I want to remove as much unnecessary steps as possible.

src/diffpy/utils/tests/test_strumining_helpers.py
sbillinge: this is how regolith does it to make a fs database backend. This is defined in conftest.py and passed into hte tests
```
@pytest.fixture(scope="session")
def make_db():
    """A test fixutre that creates and destroys a git repo in a temporary
    directory.
    This will yield the path to the repo.
    """
    cwd = os.getcwd()
    name = "regolith_fake"
    repo = os.path.join(tempfile.gettempdir(), name)
    if os.path.exists(repo):
        rmtree(repo)
    subprocess.run(["git", "init", repo])
    os.chdir(repo)
    with open("README", "w") as f:
        f.write("testing " + name)
    with open("regolithrc.json", "w") as f:
        json.dump(
            {
                "default_user_id": "sbillinge",
                "groupname": "ERGS",
                "databases": [
                    {
                        "name": "test",
                        "url": repo,
                        "public": True,
                        "path": "db",
                        "local": True,
                        "backend": "filesystem"
                    }
                ],
                "repos": [
                    {"_id": "talk_repo",
                     "params": {
                         "namespace_id": "35",
                         "initialize_with_readme": "true",
                         "name": "repo name"},
                     "url": "https://example.com",
                     "api_route": "/url/example",
                     "namespace_name": "talks"
                     }
                ],
                "stores": [
                    {
                        "name": "store",
                        "url": repo,
                        "path": repo,
                        "public": True,
                    }
                ],
                "tokens": [
                    {
                        "_id": "gitlab_private_token",
                        "token": "<private_token>"
                    }
                ],
            },
            f,
        )
    fspath = os.path.join(repo, 'db')
    os.mkdir(fspath)
    exemplars_to_fs(fspath)
    subprocess.run(["git", "add", "."])
    subprocess.run(["git", "commit", "-am", "Initial readme"])
    yield repo
    os.chdir(cwd)
    if not OUTPUT_FAKE_DB:
        rmtree(repo)
```

Sparks29032:
Tests should fail currently.

src/diffpy/utils/tests/test_strumining_helpers.py
Sparks29032: Instead of using this, pytest allows a tmp_path that is cleaned up even upon failures (confirmed with false assertions).

PR #37, opened 2023-10-09
Docs patch

Sparks29032:
Automatically publishes on release (hopefully).
[html.zip](https://github.com/diffpy/diffpy.utils/files/12847522/html.zip)


sbillinge:
This looks really good. Do you want me to merge it or y ou will put all the doc updates on this pr?  I think codecov is failing because we don't have a test for a scalar input for x

Sparks29032:
Yep, I think the codecov failures are also from a few extra line breaks I added for easier code visibility, but I didn't do any major code edits.

I think this is good to merge since I may mess some stuff up when editing the `.rst`s later on.

sbillinge:
> Yep, I think the codecov failures are also from a few extra line breaks I added for easier code visibility, but I didn't do any major code edits.
> 
> I think this is good to merge since I may mess some stuff up when editing the `.rst`s later on.

The codecov comes from line 70 in `resample.py`.  You touched this by entering a line-break so it triggered the codecov report, but it is picking up on something that should have had a test written before.  We like to take this opportunity to add missing tests if it doesn't slow us down too much.  I guess that x is usually a list but this code handles what should happen if a scalar is passed in, so a test that passes in a scalar and gives the right output is what is needed.  If it is quick it is worth doing.  It could be just one line in a `pytest.mark.parametrize` clause.  If it is more than that then we can let it slide.

sbillinge:
sweet!

PR #59, opened 2024-05-22
added eq functions for diffraction objects

yucongalicechen:


sbillinge:
I don't see any tests.  Did you write the code before the tests?

yucongalicechen:
Just added, sorry!

yucongalicechen:
I wanted to compare two default Diffraction_object(), but in the initialization function, `self.on_q = [np.empty(1), np.empty(1)] ` results in different values each time, so that Diffraction_object() != Diffraction_object(). But ideally, we want Diffraction_object() = Diffraction_object()?

sbillinge (commented):
Tests look great!!!! nice job!

In the future we may add attributes that we want to store (like datetime!) and we would like tests to fail if possible if that happens until we update the equality method to add that.  Will that happen?  Can you think a bit about that?

That said, we should think whether we want datetime (and potentially uuid) to be part of the equality.  We may want to check whether two DO's are "functionally" equal even if they were created at different times for example.  That could be way more useful.

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -21,6 +21,17 @@ def __init__(self, name='', wavelength=None):
         self._all_arrays = [self.on_q, self.on_tth]
         self.metadata = {}
 
+    def __eq__(self, other):
+        if not isinstance(other, Diffraction_object):
+            return NotImplemented
+        return (self.name == other.name and
```
sbillinge, line 27: can we think of a way of doing this by iterating over `DO.__dict__.items()`?  this way it will automatically update in the future when we add attributes.

sbillinge (commented):
please see inline comment

src/diffpy/utils/scattering_objects/diffraction_objects.py
yucongalicechen: Oh this makes sense, I'll think about it

yucongalicechen: For datetime, I think we can set `self.creation_time = datetime.now()` (we can convert it into a string if this makes everything easier), and skip certain attributes (use `reserved_keys = [creation_time]`), and I think everything else works basically the same

sbillinge (commented):
it looks good.  We may want to put in some tolerance to avoid rounding errors in the numerics for the floats.  What do you think? fewE-5?  I think most diffraction objects will be the operationally the same at the 5th decimal place.

yucongalicechen:
Yes I agree. I added the tolerance. I have one more test I need to add for comparing wavelength (didn't realize that when I committed)

sbillinge (commented):
nice, thanks.  presumably we need tests for that (pass when the numbers are different but within tolerance, fail when they are out of tolerance)?

yucongalicechen:
Yes, also when comparing float with None. I'll add both tests.

sbillinge:
@yucongalicechen maybe we could also add an asser to the tests that checks that the list of things you are testing equals DO.__dict__ too, then the test will definitely fail if something is added that is not tested?

yucongalicechen:
Sounds good, I'll add that

yucongalicechen:
I added tests and a set of attributes we want to test for, and cleaned up a bit for the tests. The function codes probablly look a bit complicated now.

sbillinge (commented):
looks good.  I think we were going to add an assert that all the attributes we want to test from DO.__dict__ are actually tested?

yucongalicechen:
- Do you mean add something like `assert self_public_attributes == attributes_to_test` so that we have gone through all attributes? I think this is probably taken care of by checking that `attributes_to_test = self_public_attributes = other_public_attributes`?
- Also, I see that in current test, `Diffraction_object() != Diffraction_object()` because np.empty() initiates to different values. I just want to make sure that it was our intention that `Diffraction_object()` are different?

sbillinge:
> * Do you mean add something like `assert self_public_attributes == attributes_to_test` so that we have gone through all attributes? I think this is probably taken care of by checking that `attributes_to_test = self_public_attributes = other_public_attributes`?
> 
>     * Also, I see that in current test, `Diffraction_object() != Diffraction_object()` because np.empty() initiates to different values. I just want to make sure that it was our intention that `Diffraction_object()` are different?

You are testing a list of attributes manually....name, wavelength, etc., so make a list of all the things you are testing and see if it is the same as the list you get from DO.__dict__ (less anything we don't want to test explicitly for some reason).  In this way, if in the future we add an attribute this test will fail until we add a test for it in `__eq__`

For your second point, no that doesn't make too much sense, we would like `Diffraction_object() == Diffraction_object()` so the test should pass.  We will have to modify the code somehow to make that pass.

yucongalicechen:
Does this work?
`        attributes_to_test = ["name", "wavelength", "scat_quantity", "on_q", "on_tth", "on_d", "metadata"]
        self_public_attributes = [key for key in self.__dict__ if not key.startswith("_")]
        other_public_attributes = [key for key in other.__dict__ if not key.startswith("_")]
        if (sorted(self_public_attributes) != sorted(other_public_attributes) or
                sorted(attributes_to_test) != sorted(self_public_attributes)):
            return False`
Or would you like an error message stating that DO has different attributes than our desired list of attributes?

sbillinge:
> Does this work?
> 
> ```
> attributes_to_test = ["name", "wavelength", "scat_quantity", "on_q", "on_tth", "on_d", "metadata"]
> self_public_attributes = [key for key in self.__dict__ if not key.startswith("_")]
> other_public_attributes = [key for key in other.__dict__ if not key.startswith("_")]
> if (sorted(self_public_attributes) != sorted(other_public_attributes) or
>     sorted(attributes_to_test) != sorted(self_public_attributes)):
>     return False
> ```
> 
> Or would you like an error message stating that DO has different attributes than our desired list of attributes?

something like this, but write it as an assert, so it is a test.  

It would also be better if the list of tested attributes could be done something like
```
attributes_to_test = ["wavelength", "name"]
for i, attribute in enumerate(attributes_to_test):
    setattr(diffraction_object1, attribute, input1[i])
    setattr(diffraction_object2, attribute, input2[i])

assert set(attributes_to_test) == set(diffraction_object1.__dict__)
```
but of course handling the private attributes and so on that we don't want in the eq method... but you get the idea.

sbillinge:
let's also check all the attributes in `DO.__dict__` and see if there are any more that we want to have in the equality

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -15,12 +15,38 @@ def __init__(self, name='', wavelength=None):
...
+        self.on_tth = [np.empty(0), np.empty(0)]
+        self.on_d = [np.empty(0), np.empty(0)]
         self._all_arrays = [self.on_q, self.on_tth]
         self.metadata = {}
 
+    def __eq__(self, other):
+        if not isinstance(other, Diffraction_object):
+            return NotImplemented
+
+        attributes_to_test = {"name", "wavelength", "scat_quantity", "on_q", "on_tth", "on_d", "metadata"}
```
sbillinge, line 28: I am not sure why this code-block is here.  It looks like the test, but we only need it as a test, not as part of the code.  I suggest to delete it.

sbillinge, line 28: since you define attributes to test here, why not import it into the test function, or vice versa?   Have a think if that makes sense.

sbillinge (commented):
This is looking great now.  Just a last couple of comment.s  Somehow one of the tests has made it into the code and it is not needed here, but only in the tests.

src/diffpy/utils/scattering_objects/diffraction_objects.py
yucongalicechen: I agree. I think it probably makes more sense to define it as a global parameter in the function and then import it to test function?

sbillinge:
Yes, sounds good

sbillinge:
come to think of it, now we figured out how to do this dynamically, why don't we do it dynamically in the function using setattr and iterating over the public attributes?  When we update DO with more attributes, we won't have to modify anything....

yucongalicechen:
Do you mean that we can put attributes_to_test just in the test function, so that later when we update DO with other attributes, we only need to add in the test, but not in the function itself?

sbillinge:
> Do you mean that we can put attributes_to_test just in the test function, so that later when we update DO with other attributes, we only need to add in the test, but not in the function itself?

That is what I was thinking.  But now I think about it it raises some interesting questions about what equality means.  If the attribute list of the DOs are different then the equality no longer commutes in principle.    DO1 == DO2  can return a different result than DO2 == DO1 because they will iterate over different lists.  We should think what we want to do here.  The most conservative take would be to check equivalence of the attribute list as part of the equality actually.....what you did.   Maybe we should leave this in though it was not my original intent....

src/diffpy/utils/tests/test_diffraction_objects.py
```diff
@@ -1,7 +1,86 @@
...
+         {"thing1": 1, "thing2": "thing2"}],
+        False,
+    ),
+]
+
+@pytest.mark.parametrize("inputs1, inputs2, expected", params)
+def test_diffraction_objects_equality(inputs1, inputs2, expected):
+    diffraction_object1 = Diffraction_object()
+    diffraction_object2 = Diffraction_object()
+    attributes_to_test = ["name", "wavelength", "scat_quantity", "on_q", "on_tth", "on_d", "metadata"]
```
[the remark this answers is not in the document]
yucongalicechen: Yeah, we should generate dynamically from the dict. Just edited!

sbillinge (commented):
this looks good now.  Please can you add a news item for this?

yucongalicechen:
I'll do that!

PR #60, opened 2024-05-22
added functions for getting username and email

yucongalicechen:
I copied everything from diffpy.labpdfproc except for the get_args, and here it returns username and email instead.

### DIFF
--- a/src/diffpy/utils/scattering_objects/tools.py
+++ b/src/diffpy/utils/scattering_objects/tools.py
@@ -1,0 +1,43 @@
+from pathlib import Path
+
+from diffpy.utils.scattering_objects.user_config import prompt_user_info, read_conf_file, write_conf_file
+
+def get_user_info():
+    """
+    Get username and email.
+
+    Prompt the user to enter username and email.
+    If not provided, read from the config file (first from cwd, and then from home directory).
+    If neither are available, raise a ValueError.
+    Save provided values to the config file if a config file doesn't exist.
+
+    Parameters
+    ----------
+    args argparse.Namespace
+        the arguments from the parser
+
+    Returns
+    -------
+    two strings on username and email
+
+    """
+    input_username, input_email = prompt_user_info()
+    conf_username, conf_email = read_conf_file()
+
+    no_username = not input_username and not conf_username
+    no_email = not input_email and not conf_email
+    if no_username and no_email:
+        raise ValueError("Please rerun the program and provide a username and email.")
+    if no_username:
+        raise ValueError("Please rerun the program and provide a username.")
+    if no_email:
+        raise ValueError("Please rerun the program and provide an email.")
+
+    username = input_username or conf_username
+    email = input_email or conf_email
+    if "@" not in email:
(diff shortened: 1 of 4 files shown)


sbillinge:
@yucongalicechen please can you let me know if slack when the tests are ready for review and I will take a look.

yucongalicechen:
Sure, I'll let you know when it's ready.

yucongalicechen:
I tested these cases:
1. No existing conf file: write valid username and email to home conf file.
2. Invalid username or email: expect error message if no valid config file or bad email input.
3. Conf file in cwd: load from input or existing cwd conf file; file remains unchanged.
4. Conf file in home directory but not cwd: load from input or existing home conf file; file remains unchanged.

I'm not too sure about the following scenarios:
5. Invalid conf file in cwd (there may be a valid conf file in home): Current code reads cwd conf file and writes inputs to home conf file (if inputs are valid).
6. Invalid conf file in home directory (no conf file in cwd): Current code writes inputs to home conf file.
7. If a conf file has only username but not email, and user inputs valid username and email, would we want to write both into conf file? (I think current code writes both to home conf file.)

sbillinge (commented):
I think the tests are still a bit too complicated.   We only need to test running in cwd and then test with 

1. a config file in cwd
2. no config file in cwd but a valid one in home_dir
3. no config file anywhere

and we only need to check the contents of the config file in `home_dir` as this is the only one that will ever be updated.   Then test the behavior when something is or isn't added through the cli.    we don't need to check paths to any files.

sbillinge:
> I tested these cases:
> 
>     1. No existing conf file: write valid username and email to home conf file.
> 
>     2. Invalid username or email: expect error message if no valid config file or bad email input.
> 
>     3. Conf file in cwd: load from input or existing cwd conf file; file remains unchanged.
> 
>     4. Conf file in home directory but not cwd: load from input or existing home conf file; file remains unchanged.
> 
> 
> I'm not too sure about the following scenarios: 5. Invalid conf file in cwd (there may be a valid conf file in home): Current code reads cwd conf file and writes inputs to home conf file (if inputs are valid). 6. Invalid conf file in home directory (no conf file in cwd): Current code writes inputs to home conf file. 7. If a conf file has only username bu not email, and user inputs valid username and email, would we want to write both into conf file? (I think current code writes both to home conf file.)

2. I am not sure this is right.  I think it should error even if it can find a valid result from a config file.  no magic.  If hte user gives a username/email at cli and it is invalid, they made a mistake and need to be told.
5. in this case error (same logic as 2., above)
6. in htis case error.  User needs to fix the conf file.
7. error.

For these errors give an error message that tells the user how to fix the issue.  We may want to have a `config.py` that runs the workflow for the user to input username and password and then write the valid conf file in home.

yucongalicechen:
Okay, I simplified the tests. I feel like it is probably less complicated when the tests are separated for the different things we are checking in each case..

sbillinge:
Yes, much clearer that way.  I don't like conditionals in tests... Makes it
very hard to follow logic

PR #67, opened 2024-05-24
added docstrings for user config

yucongalicechen:
- Added docstrings for public functions in tools.py.
- I added user_config.py in my previous PR, but it is not used, so I assume this file should be deleted?

### DIFF
--- a/src/diffpy/utils/tests/test_tools.py
+++ b/src/diffpy/utils/tests/test_tools.py
@@ -46,7 +46,7 @@ params_user_info_with_no_home_conf_file = [
     (["", ""], ["input_username", "input@email.com"], ["input_username", "input@email.com"]),
     (["cli_username", ""], ["", "input@email.com"], ["cli_username", "input@email.com"]),
     (["", "cli@email.com"], ["input_username", ""], ["input_username", "cli@email.com"]),
-    (["cli_username", "cli@email.com"], ["cli_username", "cli@email.com"], ["cli_username", "cli@email.com"]),
+    (["cli_username", "cli@email.com"], ["input_username", "input@email.com"], ["cli_username", "cli@email.com"]),
 ]
 params_user_info_no_conf_file_no_inputs = [
     ([None, None], ["", ""], ["", ""]),
@@ -54,12 +54,12 @@ params_user_info_no_conf_file_no_inputs = [
 
 
 @pytest.mark.parametrize("inputs, expected", params_user_info_with_home_conf_file)
-def test_load_user_info_with_home_conf_file(monkeypatch, inputs, expected, user_filesystem):
+def test_get_user_info_with_home_conf_file(monkeypatch, inputs, expected, user_filesystem):
     _setup_dirs(monkeypatch, user_filesystem)
     _run_tests(inputs, expected)
 
 @pytest.mark.parametrize("inputs, expected", params_user_info_with_local_conf_file)
-def test_load_user_info_with_local_conf_file(monkeypatch, inputs, expected, user_filesystem):
+def test_get_user_info_with_local_conf_file(monkeypatch, inputs, expected, user_filesystem):
     home_dir = _setup_dirs(monkeypatch, user_filesystem)
     local_config_data = {"username": "cwd_username", "email": "cwd@email.com"}
     with open(Path(user_filesystem) / "diffpyconfig.json", "w") as f:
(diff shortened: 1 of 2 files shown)


src/diffpy/utils/tests/test_tools.py
```diff
@@ -46,20 +46,20 @@ def _run_tests(inputs, expected):
     (["", ""], ["input_username", "input@email.com"], ["input_username", "input@email.com"]),
     (["cli_username", ""], ["", "input@email.com"], ["cli_username", "input@email.com"]),
     (["", "cli@email.com"], ["input_username", ""], ["input_username", "cli@email.com"]),
-    (["cli_username", "cli@email.com"], ["cli_username", "cli@email.com"], ["cli_username", "cli@email.com"]),
+    (["cli_username", "cli@email.com"], ["input_username", "input@email.com"], ["cli_username", "cli@email.com"]),
```
yucongalicechen, line 49: I think this should be input names rather than cli? And this means that cli overrides inputs?

src/diffpy/utils/tests/test_tools.py
```diff
@@ -46,20 +46,20 @@ def _run_tests(inputs, expected):
...
 ]
 params_user_info_no_conf_file_no_inputs = [
     ([None, None], ["", ""], ["", ""]),
 ]
 
 
 @pytest.mark.parametrize("inputs, expected", params_user_info_with_home_conf_file)
-def test_load_user_info_with_home_conf_file(monkeypatch, inputs, expected, user_filesystem):
+def test_get_user_info_with_home_conf_file(monkeypatch, inputs, expected, user_filesystem):
```
yucongalicechen, line 57: I changed the test names to match with the function name

src/diffpy/utils/tools.py
```diff
@@ -30,31 +68,31 @@ def _sorted_merge(*dicts):
 
 def _create_global_config(args):
     username = input(f"Please enter the name of the user to put in the diffpy global config file "
-                     f"[{args.get("username", "")}]:  ").strip() or args.get("username", "")
+                     f"[{args.get('username', '')}]:  ").strip() or args.get("username", "")
```
yucongalicechen, line 71: I changed the double quotes inside the f-string to single quotes to avoid syntax error

src/diffpy/utils/tools.py
```diff
@@ -30,31 +68,31 @@ def _sorted_merge(*dicts):
...
+    First attempts to load config file from global and local paths.
+    If neither exists, creates a global config file.
+    It prioritizes values from args, then local, then global.
+    Removes invalid global config file if creation is needed, replacing it with empty username and email.
 
     Parameters
     ----------
     args argparse.Namespace
-        the arguments from the parser
+        the arguments from the parser, default is None
```
yucongalicechen, line 91: I think I made a mistake here. The args here is a dictionary instead of a Namespace. Would we want args to be a dictionary or Namespace here?

src/diffpy/utils/tests/test_tools.py
sbillinge: no, cli overrides everything.  A user would only supply this information at runtime if they want it.   We only trigger the input sequence when there is no config file found and nothing specified on the command line for either username, or email or both.

src/diffpy/utils/tools.py
sbillinge: We could use either, but I thought dict was more general as this is, in principle, something that we want to reuse in different contexts, so I thought it was better not to expect that there was a cli in general.  When there is a cli (as in labpdfproc) we can turn the argparse.Namespace into a dict when we pass it in.

It could have been done the other way, I just made a choice here.

src/diffpy/utils/tools.py
```diff
@@ -30,31 +68,31 @@ def _sorted_merge(*dicts):
...
+    get username and email configuration
 
-    Prompt the user to enter username and email.
-    If not provided, read from the config file (first from cwd, and then from home directory).
-    If neither are available, raise a ValueError.
-    Save provided values to the config file if a config file doesn't exist.
+    First attempts to load config file from global and local paths.
+    If neither exists, creates a global config file.
+    It prioritizes values from args, then local, then global.
+    Removes invalid global config file if creation is needed, replacing it with empty username and email.
```
sbillinge, line 86: remove this last line, it is an implementational detail.

sbillinge (commented):
thanks @yucongalicechen this is great.  Please see the inline comments, then I can merge it.  Very nice.

src/diffpy/utils/tests/test_tools.py
sbillinge: oh, wait, I see what you are saying.  Yes, you are right!  Nicely done!

sbillinge:
I went ahead and merged, this was great!   I will tweak the docstrings a tiny bit on another PR so no worries about that.

I have created an issue to add this to the docs.  Would you be able to take that issue and do it soon while it is top of mind?

in the docs we want a section describing what the "utility" (in diffpy.utils!) is for.  Who would use it and why they would use it and how they can use it.  Please ask if you have any questions.

We will need to release this to conda-forge before we can really use it easily in diffpy.labpdfproc so I would like to schedule a release for say June 10th?

yucongalicechen:
I can take the issue! :)

sbillinge:
I see an "Examples" section that Andrew wrote for his two utilities.  We will need one like that.  But we also need a "Utilities" section that just has a few words for each utility (one-to-one correspondence with the ones in the Examples ection).  These don't have to be long.  Just what the utility does and who might use and for what.  Just do your best (@Sparks29032 could maybe contribute these for his utilities) and I can tweak later.

PR #73, opened 2024-05-30
add function to get time and package information

yucongalicechen:
- Initial commit
- added function to get package version, loaded time and package into DO.metadata through dump


sbillinge (commented):
I thought I wrote a review but I can't find it, so in case you didn't get it I summarize here again.

This is very nice but I think the design of this could maybe be tweaked a bit.  I think a tool that can extract the package information and return it in a dict format would be widely useful so could be in diffpy.utils.tools.  maybe takes a package name and optionally a metada dict and returns the metadata dict updated, something like {"package_info": `[("package1, 1.2.3), ("package2", "2.3.7")]}`

Then diffpy.utils.diffractionobejcts can take "package_info" if it finds it and add itself when it is instantiated by calling the tool?  We can decide if we need to "pretty print" it in dump but just try dumping it to begin with and see how human readable it is maybe.

I think this would make it more reusable.

yucongalicechen:
I agree that this makes more sense as it addresses the cases when there is no diffraction object.

The `get_package_info` function currently inserts or updates the package_info in the given metadata. I'm wondering if it makes more sense to let the function extends the package_info if it already exists?
i.e. `get_package_info("package1", {"package_info": [("package1", "1.2.3"), ("package2", "2.3.7")]})`
will return `{"package_info": [("package1", "1.2.4"), ("package2", "2.3.7")]}`,
and `get_package_info("package3", {"package_info": [("package1", "1.2.3"), ("package2", "2.3.7")]})`
will return `{"package_info": [("package1", "1.2.3"), ("package2", "2.3.7"), ("package3", "3.4.5")]}`.

I let the dump function call "get_package_info" automatically if it cannot find package_info in the current metadata or the user specifies a package_name, not sure if this is the desired behavior.

sbillinge:
> I agree that this makes more sense as it addresses the cases when there is no diffraction object.
> 
> The `get_package_info` function currently inserts or updates the package_info in the given metadata. I'm wondering if it makes more sense to let the function extends the package_info if it already exists? i.e. `get_package_info("package1", {"package_info": [("package1", "1.2.3"), ("package2", "2.3.7")]})` will return `{"package_info": [("package1", "1.2.4"), ("package2", "2.3.7")]}`, and `get_package_info("package3", {"package_info": [("package1", "1.2.3"), ("package2", "2.3.7")]})` will return `{"package_info": [("package1", "1.2.3"), ("package2", "2.3.7"), ("package3", "3.4.5")]}`.
> 
> I let the dump function call "get_package_info" automatically if it cannot find package_info in the current metadata or the user specifies a package_name, not sure if this is the desired behavior.

yes, this is the behavior we want.  takes a package name (string) and a metadata dict, looks for `package_info` in the dict and extends it, or creates it if it can't find it.    This is simple, easy to maintain, and will satisfy all the UCs.

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -456,7 +460,7 @@ def on_xtype(self, xtype):
             return self.on_d
         pass
 
-    def dump(self, filepath, xtype=None):
+    def dump(self, filepath, xtype=None, package_name=None):
```
sbillinge, line 463: this update is not needed.

src/diffpy/utils/tests/test_diffraction_objects.py
```diff
@@ -230,26 +233,70 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
-        "[Diffraction_object]\nname = test\nwavelength = 1.54\nscat_quantity = x-ray\nthing1 = 1\n"
-        "thing2 = thing2\n\n#### start data\n0.000000000000000000e+00 0.000000000000000000e+00\n"
-        "1.000000000000000000e+00 1.000000000000000000e+00\n2.000000000000000000e+00 2.000000000000000000e+00\n"
-        "3.000000000000000000e+00 3.000000000000000000e+00\n4.000000000000000000e+00 4.000000000000000000e+00\n"
-        "5.000000000000000000e+00 5.000000000000000000e+00\n"
-        "6.000000000000000000e+00 6.000000000000000000e+00\n7.000000000000000000e+00 7.000000000000000000e+00\n"
-        "8.000000000000000000e+00 8.000000000000000000e+00\n9.000000000000000000e+00 9.000000000000000000e+00\n"
-        "1.000000000000000000e+01 1.000000000000000000e+01\n"
+def test_dump(tmp_path, monkeypatch):
+    monkeypatch.setattr(
```
sbillinge, line 237: Is there a way to use monkeypatch for everything (not sure but it would be easier to read if you can)

src/diffpy/utils/tests/test_diffraction_objects.py
```diff
@@ -230,26 +233,70 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+        importlib.metadata, "version", lambda package_name: "3.3.0" if package_name == "diffpy.utils" else "1.2.3"
     )
-    assert actual == expected
+    with patch("diffpy.utils.scattering_objects.diffraction_objects.datetime") as mock_datetime:
+        mock_datetime.now.return_value = datetime(2024, 5, 30, 12, 30, 1)
```
sbillinge, line 241: see below, but I think it is easier to read if there is less in the context manager.

sbillinge:
Also, please don't close this PR, but for future reference, there is a bit too much going on in it.  It is easier to review and merge if you separate the tools work from the dump work into separate PRs.   I know that one uses the other, but we can start working on the tests without the tools merged, and That part of the PR is almost ready for merge anyway, and then you can merge main into the dump branch to finish up.

yucongalicechen:
(my previous commit message disappeared when I was merging conflicts)
I edited package_info function (to extend package_info instead of updating completely) and simplified dump and the tests.
I'll add to the news, I assume I also need to update the docs for tools?

sbillinge:
@yucongalicechen I am closing this issue.  I merged all the changes on two other PRs.  this one had main merged in and I would rather do it the other way around (basically never merge main into a branch unless it is necessary because you need some functionality or there is a conflict that can't be undone locally).  

I also cleaned up the functionality a little bit.  The main things were to make the `package_info` value a dict instead of a list of tuples.  I had suggested the latter, but it was way too nested to do checking easily so it was a bad idea! :).  I also cleaned up the mocking to make it more readable in the tests.  there were a few other small tweaks.  You may want to take a look if you like.

PR #75, opened 2024-05-31
Edits made by running pre-commit

cadenmyers13:
excludes edits made to rst and txt files


sbillinge:
This is great, thanks @cadenmyers13 

This may change a bit with the new pre-commit of Andrew but probably not much if at all so happy to merge this.

PR #78, opened 2024-06-03
Flake8 fixes (2/3)

cadenmyers13:
Similar edits to other Flake8 PR

### DIFF
--- a/src/diffpy/utils/tests/test_serialization.py
+++ b/src/diffpy/utils/tests/test_serialization.py
@@ -93,3 +93,18 @@ def test_exceptions():
     no_dt = serialize_data(nodt, nodt_hdata, nodt_dt, show_path=False)
     nodt_data_name = list(no_dt.keys())[0]
     assert numpy.allclose(no_dt[nodt_data_name]["data table"], nodt_dt)
+
+    # ensure user is warned when columns are overwritten
+    hdata = loadData(warningfile, headers=True)
+    data_table = loadData(warningfile)
+    with pytest.warns(RuntimeWarning) as record:
+        serialize_data(
+            warningfile,
+            hdata,
+            data_table,
+            show_path=False,
+            dt_colnames=["c1", "c2", "c3"],
+        )
+    assert len(record) == 4
+    for msg in record:
+        assert "overwritten" in msg.message.args[0]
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/tests/test_serialization.py
```diff
@@ -93,18 +93,3 @@ def test_exceptions():
...
-        data = serialize_data(
-            warningfile,
-            hdata,
-            data_table,
-            show_path=False,
-            dt_colnames=["c1", "c2", "c3"],
-        )
-    assert len(record) == 4
-    for msg in record:
-        assert "overwritten" in msg.message.args[0]
```
cadenmyers13, line 110: Discussed with Andrew about the removal of unused variables. Since testing is done I removed this section.

src/diffpy/utils/tests/test_tools.py
```diff
@@ -91,7 +91,6 @@ def test_get_user_info_with_home_conf_file(monkeypatch, inputs, expected, user_f
 
 @pytest.mark.parametrize("inputs, expected", params_user_info_with_local_conf_file)
 def test_get_user_info_with_local_conf_file(monkeypatch, inputs, expected, user_filesystem):
-    home_dir = _setup_dirs(monkeypatch, user_filesystem)
```
cadenmyers13, line 95: Similarly, removed this because testing is done.

src/diffpy/utils/tests/test_serialization.py
Sparks29032: Wait I meant remove the `data` variable. Keep everything else. Just remove `data = `.

src/diffpy/utils/tests/test_tools.py
```diff
@@ -92,7 +91,6 @@ def test_get_user_info_with_home_conf_file(monkeypatch, inputs, expected, user_f
...
```
Sparks29032, line 95: home_dir might be used by `pytest`

Sparks29032:
CI errors seem to result from either removing `argparse` or `home_dir`. Not sure which one.

src/diffpy/utils/tests/test_serialization.py
sbillinge: Don't ever reomove tests.  Testing is never "done", we leave the tests there forever and they protect against breaking changes in the future.

src/diffpy/utils/tests/test_tools.py
sbillinge: don't remove

sbillinge (commented):
some misunderstanding here.  Please see comments.

PR #89, opened 2024-06-18
Remove `labpdfproc` subdirectory from git ignore

Sparks29032:
Maybe a residue from testing? Either way, does not seem like it should be here.


sbillinge:
thanks for catching that.

PR #95, opened 2024-09-07
Migrate from unittest to pytest

bobleesj:
Before migration: 43 passed and 2 warnings 
After migration: 43 passed and 2 warnings 

Remove files such as `run.py` and `debug.py` that were used to collect and run unittest functions.


bobleesj:
GH CI results from own fork:

```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================== 43 passed, 2 warnings in 0.59s ========================
Name                                                         Stmts   Miss  Cover   Missing
```

https://github.com/bobleesj/diffpy.utils/actions/runs/10751954274/job/29819744369

sbillinge (commented):
We don't have to remove unittest tests. Pytest will run unittest tests. Moving to pytest just means that future tests can be pytest tests.

sbillinge:
Let's redo this PR keeping all the unittest tests but removing the run.py etc. And making sure the tests all pass.  Then we can move the tests directory to the to level and test again?  What do you think?

bobleesj:
Okay, sounds good. It may not be practical to migrate from unittest to pytest for other packages.

The main priority is to separate test helper function(s) that exist within the `src` directory such as

```
from diffpy.utils.tests.testhelpers import datafile
```

into `tests/conftest.py`. I have done it in another branch but I will try to keep using unittest and try to remove run files.

sbillinge:
Good.  Is better to keep tests to be sure that whatever was tested is still tested which is why I would rather revert that work (assuming the original tests were good tests)

PR #96, opened 2024-09-07
Move tests folder to top and use pytest to run all tests 

bobleesj:
43 tests are passing, same as before and after moving the `tests`.

Run via: 

```
pytest

...

43 passed, 2 warnings in 0.43s 
```


tests/test_loaddata.py
```diff
@@ -6,18 +6,20 @@
...
 from diffpy.utils.parsers import loadData
-from diffpy.utils.tests.testhelpers import datafile
-
-loaddata01 = datafile("loaddata01.txt")
-loaddatawithheaders = datafile("loaddatawithheaders.txt")
 
 
 ##############################################################################
 class TestLoadData(unittest.TestCase):
+    @pytest.fixture(autouse=True)
```
bobleesj, line 16: Trick to use `@pytest.fixture` within `unittest`

Source: https://stackoverflow.com/a/66222108

pytest.ini
```diff
@@ -1,3 +1 @@
 [pytest]
-testpaths= src/diffpy/utils/tests
-python_files = test_*.py
```
bobleesj, line 3: pytest automatically detects files starting with `test_.`

bobleesj (commented):
Please review!

bobleesj:
```pytest arguments: ['-v']
============================= test session starts ==============================

plugins: mock-3.14.0, env-1.1.3
collecting ... collected 43 items

tests/test_diffraction_objects.py::test_diffraction_objects_equality[inputs10-inputs20-True] PASSED [  2%]

...

tests/test_tools.py::test_get_package_info[inputs3-expected3] PASSED     [100%]

=============================== warnings summary ===============================
```

Seems like the tests are passing while we probably need to codecov and other CI files.

Maybe shall we merge it a new branch like "folder" etc.? @sbillinge

sbillinge (commented):
The tests seem to still be in the src package.  We want them all in tests at the top level, no? 

When you move them please use `git mv` so got just updates the path and doesn't delete and add the whole file.

bobleesj:
Screenshot 2024-09-07 at 4 27 58 PM

@sbillinge I think the test files have removed via `git mv` already.  Could you please double-check?

sbillinge:
@bobleesj this looks great, thanks.  Sorry, before I was looking on my phone and didn't see the right-arrow showing the file move so I missed that it was all good.

PR #98, opened 2024-09-08
Remove older files outside of cookiecutter template

bobleesj:
My understanding is that they are no longer used.


bobleesj:
The test fails because `run_test.py` has removed but the older CI version here uses it to run tests.

Once https://github.com/diffpy/diffpy.utils/pull/97 is merged, this will most likely to pass.

bobleesj:
Re-run CI

bobleesj:
@sbillinge Good to go! Please review

PR #112, opened 2024-09-18
Remove two Pytest warnings due to numpy and pytest mocker in test_dump function

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/133


tests/test_diffraction_objects.py
```diff
@@ -236,18 +236,21 @@ def test_dump(tmp_path, mocker):
     directory = Path(tmp_path)
     file = directory / "testfile"
     test = Diffraction_object()
-    test.wavelength = 1.54
+    test.wavelength = 1.01
```
bobleesj, line 239: The test function is designed to log output data using arbitrary numbers. Using a value of 1.54 led to a runtime warning due to an issue with the arcsin function in NumPy, which expects inputs between -1 and 1. 

When the input exceeds this range, it produces NaN outputs and a `RuntimeWarning`

It is now resolved.

tests/test_diffraction_objects.py
```diff
@@ -236,18 +236,21 @@ def test_dump(tmp_path, mocker):
...
     test.name = "test"
     test.scat_quantity = "x-ray"
     test.insert_scattering_quantity(
         x, y, "q", metadata={"thing1": 1, "thing2": "thing2", "package_info": {"package2": "3.4.5"}}
     )
-    with mocker.patch("importlib.metadata.version", return_value="3.3.0"), freeze_time("2012-01-14"):
+
+    mocker.patch("importlib.metadata.version", return_value="3.3.0")
```
bobleesj, line 246: The second warning is resolved by not using `with`  because the mocker fixture automatically handles the setup and teardown of mocks without needing manual intervention.

bobleesj (commented):
Ready for review

tests/test_diffraction_objects.py
sbillinge: This doesn't seem right.  Remember, we don't change the test to make it pass, we make the test to capture the behavior we want, then we write the code so the tests pass.  I am not sure I understand, but are you saying that the code should not work for a wavelength of 1.54?  That is your desired behavior?

sbillinge (commented):
please see inline

tests/test_diffraction_objects.py
bobleesj: The test passes regardless of wavelength = 1.54 or 1.01. Both 1.54 and 1.01 are arbitrary values.

But if 1.54 is used, the following warning (test still passes) arises:

```
/Users/imac/miniconda3/envs/diffpy.utils_env/lib/python3.12/site-packages/diffpy/utils/scattering_objects/diffraction_objects.py:324: RuntimeWarning: invalid value encountered in arcsin
    return np.rad2deg(2.0 * np.arcsin(q * pre_factor))
```

When an arbitrary value of 1.01 is used, then we have no warning because 

```
        return np.rad2deg(2.0 * np.arcsin(q * pre_factor))

Comment (bobleesj) [reply tests/test_diffraction_objects.py:239]: The test passes regardless of wavelength = 1.54 or 1.01. Both 1.54 and 1.01 are arbitrary values. 

The test function basically takes an arbitrary wavelength value and logs the outputs which is basically either 1.54 or 1.01 depending on the wavelength input.

But if 1.54 is used, the following warning (test still passes) arises:

```
/Users/imac/miniconda3/envs/diffpy.utils_env/lib/python3.12/site-packages/diffpy/utils/scattering_objects/diffraction_objects.py:324: RuntimeWarning: invalid value encountered in arcsin
    return np.rad2deg(2.0 * np.arcsin(q * pre_factor))
```

because the arbitrary value of 1.54 is used to to find an angle in the class:

```
        return np.rad2deg(2.0 * np.arcsin(q * pre_factor))
```

When an arbitrary value of 1.01 is used, then we have no warning. 

I hope I am making sense here.

Comment (sbillinge) [reply tests/test_diffraction_objects.py:239]: Yes, but this is not my point. My point is what behavior are we testing with this test? Not what code are we testing but what behavior?

**Comment by bobleesj** [comment]
> Yes, but this is not my point. My point is what behavior are we testing with this test? Not what code are we testing but what behavior?

I hope I get this right - the software behavior is expected to return a Runtime warning message from the input of 1.54. So we capture that warning message? @sbillinge

File: tests/test_diffraction_objects.py [hunk 3]
```diff
@@ -239,11 +240,29 @@ def test_dump(tmp_path, mocker):
...
+
+    with warnings.catch_warnings(record=True) as captured_warnings:
+        # Configure the warnings system to capture all warnings
+        warnings.simplefilter("always")
+
+        # Perform the method call that is expected to trigger the RuntimeWarning due to the specified wavelength
+        test.insert_scattering_quantity(
+            x, y, "q", metadata={"thing1": 1, "thing2": "thing2", "package_info": {"package2": "3.4.5"}}
+        )
+
```
bobleesj, line 252: The code now captures the warning:

From:
Screenshot 2024-09-19 at 7 26 55 AM

to: 

Screenshot 2024-09-19 at 7 27 35 AM

sbillinge:
> > Yes, but this is not my point. My point is what behavior are we testing with this test? Not what code are we testing but what behavior?
> 
> I hope I get this right - the software behavior is expected to return a Runtime warning message from the input of 1.54. So we capture that warning message? @sbillinge

I see.  You want the code to return a warning if the user inputs a wavelength of 1.54?  Why do you want the code to do that?  Is 1.54 somehow invalid as a wavelength?  Actually, 1.54 is Cu K_alpha wavelength.  It is one of the most commonly used x-ray wavelengths in the world.  I think it is a perfectly valid wavelength, or am I missing something.

tests/test_diffraction_objects.py
```diff
@@ -232,18 +233,22 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
 
 
 def test_dump(tmp_path, mocker):
-    x, y = np.linspace(0, 10, 11), np.linspace(0, 10, 11)
+    x, y = np.linspace(0, 5, 6), np.linspace(0, 5, 6)
```
bobleesj, line 236: To recollect our memory: we had this invalid value warning:

```
RuntimeWarning: invalid value encountered in arcsin
    print(np.rad2deg(2.0 * np.arcsin(q * pre_factor)))
```

**Problem:**

In our dummy `test_dump` function: 

The code sets arbitrary q values as:

```
# x, y = np.linspace(0, 10, 11), np.linspace(0, 10, 11) sets q
# q = np.asarray(q)
# print(q)
[ 0.  1.  2.  3.  4.  5.  6.  7.  8.  9. 10.]
```

Hence, when we try to convert q to two-theta, using `test.wavelength = 1.54`:

```
print(np.rad2deg(2.0 * np.arcsin(q * pre_factor)))
[  0.          14.07850645  28.37532297  43.14126254  58.70709081
  75.57672216  94.66443973 118.15097974 157.27157928          nan
          nan]
```

We have `nan` values.

**Solution:**

Just simply use the valid q range from 0 to 6, etc.

bobleesj (commented):
@sbillinge ready for review - once merged. we can do 3.12 (recut) pypi release

bobleesj:
@sbillinge applied pre-commit - ready for review

Issue #140, opened 2024-10-26
better wording on the capture user names functionality


sbillinge:
### Problem

user-name metadata capture utility is not performing as hoped.

### Proposed solution

We want to be able to make it easy to capture user metadata which is why we have the workflow for capturing the user metadata but the user messages are resulting in behavior where we either don't capture the name or capture dummy information that then gets stored in a user variable for all time.  We need to nudge users to capture accurate data so that they can get credit for their work in the future.

bobleesj:
@alisnwu

Providing a bit more context: https://github.com/diffpy/diffpy.labpdfproc/issues/118#issuecomment-2439743613 and the code is used here: https://github.com/diffpy/diffpy.labpdfproc/blob/666a5ebd96d42f0cf8837f797db5fdeb47e2d391/src/diffpy/labpdfproc/tools.py#L5

bobleesj:
closed by the PR above

bobleesj:
@sbillinge could you plz close this?

PR #149, opened 2024-11-05
Added extra information in warning for better capturing of user info

### LINKED ISSUE #129: Add instructions/link to fix missing user info issue in get_user_info function in tools.py
### Problem

Instructions for getting user info can be clearer. Right now there are no instructions in the warning message on how to fix missing user info.

### Proposed solution
Edit messages in `get_user_info` function in `tools.py`. Specifically, include instructions in the warning message and/or include a link to the right place in the online docs for fixing missing user info issue.
For more details, see discussions in https://github.com/diffpy/diffpy.labpdfproc/issues/113.

alisnwu:
closes #129 and #140

@sbillinge ready for review


news/get-user-fix.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
```
sbillinge, line 1: we can delete this file.

src/diffpy/utils/tools.py
```diff
@@ -114,6 +115,11 @@ def get_user_info(args=None):
     global_config = load_config(Path().home() / "diffpyconfig.json")
     local_config = load_config(Path().cwd() / "diffpyconfig.json")
     if global_config is None and local_config is None:
+        warnings.warn(
```
sbillinge, line 118: This is not quite what I had in mind.  Just the warning messages in the prompts could be clearer.  We could put this preamble message here if we really like it, but I would like to see it in the context of the full workflow pasted in the comments before deciding.  

@alisnwu have you run through the whole workflow to see what it looks like and where it can be improved?

Even if we do want this wording, I think this is not a good use of `warnings.warn`.  We are not using any of its features. A simple print statement would be clearer.

sbillinge (commented):
maybe the best way to do this is to paste the outputs into the comments under different situations and we can discuss what is the clearest wording?

src/diffpy/utils/tools.py
sbillinge: yes, I like that (modulo a slight grammatical boo boo)!

How about something a bit more human friendly:
"No global configuration file was found containing information about the user to associate with the data.  By following the prompts below you can add your name and email to this file on the current computer and your name will be automatically associated with subsequent diffpy data by default.  This is not recommended on a shared or public computer.  You will only have to do that once.  For more information <as before>"

we need a workflow for people to update this global config file later if they want.  Pease could you make an issue to check/work on that and maybe do it on a separate PR?  Thanks!

alisnwu:
Screenshot 2024-11-06 at 6 32 37 PM

@sbillinge ready for review

sbillinge:
Thanks @alisnwu 

It seems to be failing pre-commit.  Also, I think it would be easier to read if it were formatted a bit better on the terminal.  This will depend on the configuration on the user's computer but having some line-breaks and/or empty lines to make it easier to read would be nice I think?  What do you think?

Please can you check with Bob why the tests are not running?

bobleesj:
@alisnwu  Once you are done with formatting, could you tag me? Let's see if the tests are running. Also, plz don't forget to run `pre-commit run --all-files`! Feel free to configure a shortcut/alias like `pc` so save time - This habit saves time for the reviewers and also makes the history clean.

Ref: https://github.com/bobleesj/command-line-cheatsheet/?tab=readme-ov-file#command-line-shortcuts

alisnwu:
Screenshot 2024-11-06 at 6 51 46 PM

@sbillinge I added some newline to make the message easier to read, how does this look?

bobleesj:
> Please can you check with Bob why the tests are not running?

@sbillinge https://github.com/diffpy/diffpy.utils/pull/159 CI Pytest running again - need to merge this to see the global effect

sbillinge:
nice!  Thanks @alisnwu !

PR #154, opened 2024-11-06
add functions to allow conversions between d and tth, and d and q

### LINKED ISSUE #146: diffraction objects implement d-tth and d-q converter functions
### Problem

We don’t have a converter between d and tth, or between d and q.

### Proposed solution

Add functions d_to_tth, d_to_q, tth_to_d, q_to_d in diffraction_objects. d = 2*pi/Q.
Refer to https://github.com/diffpy/diffpy.labpdfproc/pull/120 for more details.

yucongalicechen:
closes #146 

Added functions to convert between tth and d, and d and q. Modified the function set_all_arrays so that when we create a new diffraction object, we have arrays on all xtypes. Added tests for these modified functions.

@sbillinge ready for review

### DIFF
--- a/src/diffpy/utils/scattering_objects/diffraction_objects.py
+++ b/src/diffpy/utils/scattering_objects/diffraction_objects.py
@@ -283,7 +283,7 @@ class Diffraction_object:
         elif xtype.lower() in ANGLEQUANTITIES:
             self.on_tth = [np.array(xarray), np.array(yarray)]
         elif xtype.lower() in DQUANTITIES:
-            self.on_tth = [np.array(xarray), np.array(yarray)]
+            self.on_d = [np.array(xarray), np.array(yarray)]
         self.set_all_arrays()
 
     def q_to_tth(self):
--- a/src/diffpy/utils/scattering_objects/diffraction_objects.py
+++ b/src/diffpy/utils/scattering_objects/diffraction_objects.py
@@ -323,9 +323,11 @@ class Diffraction_object:
         q = np.asarray(q)
         wavelength = float(self.wavelength)
         pre_factor = wavelength / (4 * np.pi)
-        # limit q * pre_factor to 1 to avoid invalid input to arcsin
-        arcsin_value = np.where(q * pre_factor <= 1, q * pre_factor, 1)
-        return np.rad2deg(2.0 * np.arcsin(arcsin_value))
+        if np.any(np.abs(q * pre_factor) > 1):
+            raise ValueError(
+                "Invalid input for arcsin: some values exceed the range [-1, 1]. " "Check wavelength or q values."
+            )
+        return np.rad2deg(2.0 * np.arcsin(q * pre_factor))
 
     def tth_to_q(self):
         r"""
@@ -362,6 +364,8 @@ class Diffraction_object:
             of ``wavelength``
         """
         two_theta = np.asarray(np.deg2rad(self.on_tth[0]))
+        if np.any(two_theta > np.pi):
+            raise ValueError("Two theta exceeds 180 degrees.")
         wavelength = float(self.wavelength)
         pre_factor = (4 * np.pi) / wavelength
         return pre_factor * np.sin(two_theta / 2)
@@ -425,6 +429,8 @@ class Diffraction_object:
             of ``wavelength``
         """
         two_theta = np.asarray(np.deg2rad(self.on_tth[0]))
+        if np.any(two_theta > np.pi):
+            raise ValueError("Two theta exceeds 180 degrees.")
         wavelength = float(self.wavelength)
         sin_two_theta = np.sin(two_theta / 2)
         d = np.where(sin_two_theta != 0, wavelength / (2 * sin_two_theta), np.inf)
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -362,18 +362,125 @@ def tth_to_q(self):
...
+            An array of :math:`d` values
+
+        Returns
+        -------
+        q : array
+            An array of :math:`q` values in the inverse of the units of ``wavelength``
+        """
+        epsilon = 1e-10
+        d = np.asarray(self.on_d[0])
+        d = np.where(np.abs(d) < epsilon, d + epsilon, d)
```
yucongalicechen, line 404: Added d to a small number to avoid dividing by 0, which would cause q to become some arbitrary large number (as commented later in the test function). Not sure if this is a good way to address this problem - do we want to just raise error in these cases?

tests/test_diffraction_objects.py
```diff
@@ -231,6 +231,85 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+    expected_q = [0, 4.58087]
+    assert np.allclose(actual_q, expected_q)
+
+
+def test_q_to_d():
+    actual = Diffraction_object(wavelength=0.71)
+    setattr(actual, "on_q", [[0, 4.58087], [1, 1]])
+    actual_d = actual.q_to_d()
+    expected_d = [62831853071.8, 1.37161]
+    assert np.allclose(actual_d, expected_d)
```
yucongalicechen, line 255: test case for q=0, d is an arbitrary large number

yucongalicechen:
@sbillinge I edited the codes based on our previous conversation, it's ready for review. The history for this PR is getting a bit messy, and I also saw Bob's PR for change the name for Diffraction object -- I can probably start a new PR for this after his PR is merged?
Anyway, I defined DMAX and QMAX and edited the test cases. One other issue is that is we have tth > 180, the q and d we get will "snapback" and become incorrect. Shall I set a limit for tth so that whenever tth > 180, we force tth to 180?

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -315,13 +317,15 @@ def q_to_tth(self):
...
         two_theta : array
-            An array of :math:`2\theta` values in radians
+            The array of :math:`2\theta` values in radians
         """
         q = self.on_q[0]
         q = np.asarray(q)
         wavelength = float(self.wavelength)
         pre_factor = wavelength / (4 * np.pi)
-        return np.rad2deg(2.0 * np.arcsin(q * pre_factor))
+        # limit q * pre_factor to 1 to avoid invalid input to arcsin
```
sbillinge, line 326: I am not sure about this.  If arcsin_value is greater than 1 something bad has gone wrong.  We probably want to raise an exception and let the user know.  It probably means they have specified a wrong wavelength of something.

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -364,99 +368,92 @@ def tth_to_q(self):
...
         d : array
-            An array of :math:`d` values in the inverse of the units of ``wavelength``
+            The array of :math:`d` values in the inverse of the units of ``wavelength``
         """
-        epsilon = 1e-10
         q = np.asarray(self.on_q[0])
-        q = np.where(np.abs(q) < epsilon, q + epsilon, q)
-        return (2 * np.pi) / q
+        d = np.where(q != 0, (2 * np.pi) / q, np.inf)
+        d = np.minimum(d, DMAX)
```
sbillinge, line 385: This is not very readable.  Also, I guess it can end up with a bunch of values of D that are the same.  I think there is a better way to do this maybe.  Also, how does the user overrid DMAX if they want?

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -10,6 +10,8 @@
 DQUANTITIES = ["d", "dspace"]
 XQUANTITIES = ANGLEQUANTITIES + DQUANTITIES + QQUANTITIES
 XUNITS = ["degrees", "radians", "rad", "deg", "inv_angs", "inv_nm", "nm-1", "A-1"]
+QMAX = 40
```
sbillinge, line 13: this is too low probably.

tests/test_diffraction_objects.py
```diff
@@ -232,76 +232,86 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
 
 
 def test_q_to_tth():
-    actual = Diffraction_object(wavelength=0.71)
-    setattr(actual, "on_q", [[0, 4.58087], [1, 1]])
+    actual = Diffraction_object(wavelength=4 * np.pi)
+    setattr(actual, "on_q", [[0, 0.2, 0.4, 0.6, 0.8, 1, 40, 60], [1, 2, 3, 4, 5, 6, 7, 8]])
     actual_tth = actual.q_to_tth()
-    expected_tth = [0, 30]
+    # expected tth values are 2 * arcsin(q)
```
sbillinge, line 238: this also doesn't make any sense to me.  I don't think we want this behavior.

sbillinge (commented):
Let's focus on the tests and iterate those so we get the behavior we want.  Basically if Q goes out of bounds (tth > 180) something has gone wrong and I htink we should raise and exception alerting this to user and asking if the wavelength is correctly specified.

src/diffpy/utils/scattering_objects/diffraction_objects.py
yucongalicechen: The user can use insert_scattering_quantity to overrid or directly set the attributes. But I agree with you that this might be problematic. Also the appropriate limit seems a bit different for different wavelength. Maybe it's better to set d = 2 * previous_d if q = 0?
e.g. q = [0, 1] => d = [2pi, 4pi]
q = [0, 2] => d = [pi, 2pi]

yucongalicechen:
no need to review - just for saving progress

sbillinge:
@yucongalicechen I will close this since it is replaced by those issues (and future PRs).  Thanks for doing this.

PR #155, opened 2024-11-06
Moved resample out of parsers to its own sub-package

### LINKED ISSUE #138: move resamplers out of parsers?
### Problem

### Proposed solution

alisnwu:
closes #138 

Screenshot 2024-11-06 at 5 50 22 PM

@sbillinge ready for review


src/diffpy/utils/parsers/__init__.py
```diff
@@ -17,11 +17,10 @@
 """
 
 from .loaddata import loadData
-from .resample import resample
 from .serialization import deserialize_data, serialize_data
 
 # silence the pyflakes syntax checker
-assert loadData or resample or True
+assert loadData or True
```
bobleesj, line 23: I am afraid this isn't the latest cookiecutting standard for handling `__init__.py` files within `src`. Could you please check whether we have to import these functions lilke `from .loaddata import loadData`?

bobleesj:
@alisnwu Thanks for this quick fix. For future PRs, since pytest is running via GitHub CI, we don't have to attach the screenshot but appreciate the idea.

src/diffpy/utils/parsers/__init__.py
alisnwu: No we don't, should I delete these then?

sbillinge:
@alisnwu this sseems to have picked up a conflict

Issue #157, opened 2024-11-07
Fix Test on PR GitHub Pytest CI not running


bobleesj:
### Problem

- https://github.com/diffpy/diffpy.utils/pull/154
- https://github.com/diffpy/diffpy.utils/pull/149

### Proposed solution

Investigate and leave a comment below

bobleesj:
## Solution

Error message:

[Invalid workflow file: .github/workflows/tests-on-pr.yml#L12](https://github.com/bobleesj/diffpy.utils/actions/runs/11713975461/workflow)
The workflow is not valid. .github/workflows/tests-on-pr.yml (Line: 12, Col: 11): Secret CODECOV_TOKEN is required, but not provided while calling.

> We need to provide codecov secret for the `tests-on-pr.yml`  (new cookiecutting standard)

bobleesj:
I will fix it.

PR #159, opened 2024-11-07
Add Codecov secret to `.github/workflows/tests-on-pr.yml` to use Codecov

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/151
Closes https://github.com/diffpy/diffpy.utils/issues/157


bobleesj:
@sbillinge Pytest CI running now. No news added.

PR #163, opened 2024-11-08
Move resampler out of parsers, remove relative paths for import

### LINKED ISSUE #138: move resamplers out of parsers?

alisnwu:
closes #138 

All discussions made in #158 

@sbillinge Ready for review


src/diffpy/utils/parsers/data_loader.py
```diff
@@ -16,7 +16,7 @@
 import numpy
 
 
-def loadData(filename, minrows=10, headers=False, hdel="=", hignore=None, **kwargs):
+def load_data(filename, minrows=10, headers=False, hdel="=", hignore=None, **kwargs):
```
sbillinge, line 19: we do not want this change (please see the discussion on #163 ).  Please revert.

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -17,7 +17,7 @@
 )
 
 
-class Diffraction_object:
+class DiffractionObject:
```
sbillinge, line 20: we want this change, but we want to do it as a deprecation.  Please see the discussion on #163

sbillinge (commented):
this has gone a bit wrong again.  Let's close this and try again.

sbillinge:
@alisnwu I will close this. I was expecting you to do it this way, but we don't want the committee and the reverts in the history.  Sorry for the wasted work of reverting commits

PR #165, opened 2024-11-09
Deprecate `Diffraction_object`, rename to `DiffractionObject` 

bobleesj:


src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -18,6 +19,20 @@
 
 
 class Diffraction_object:
+    """FIXME: Add class docstring.
+
+    .. deprecated:: 3.5.1
```
bobleesj, line 24: Following the numpy deprecation standards:

Screenshot 2024-11-09 at 12 28 11 PM

Ref: https://numpydoc.readthedocs.io/en/latest/format.html#deprecation-warning

bobleesj (commented):
Ready for review - learning how to handle deprecation more gracefully in our group.

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -18,6 +19,20 @@
...
+        `Diffraction_object` will be removed in diffpy.utils 3.6.0, it is replaced by
+        `DiffractionObject` to follow the class naming convention.
+    """
+
+    warnings.warn(
```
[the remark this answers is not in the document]
bobleesj: Fixed

src/diffpy/utils/scattering_objects/diffraction_object.py
```diff
@@ -0,0 +1,492 @@
...
+XQUANTITIES = ANGLEQUANTITIES + DQUANTITIES + QQUANTITIES
+XUNITS = ["degrees", "radians", "rad", "deg", "inv_angs", "inv_nm", "nm-1", "A-1"]
+
+x_grid_emsg = (
+    "objects are not on the same x-grid. You may add them using the self.add method "
+    "and specifying how to handle the mismatch."
+)
+
+
+class DiffractionObject:
```
bobleesj, line 20: A new file for `DiffractionObject`. File name changed from `diffraction_objects` to `diffraction_object`

tests/test_diffraction_object.py
```diff
@@ -0,0 +1,265 @@
+from pathlib import Path
```
bobleesj, line 1: Same comment above - new file since file name changed

src/diffpy/utils/scattering_objects/diffraction_object.py
```diff
@@ -0,0 +1,492 @@
...
+    """
```
bobleesj, line 21: I included docstrings for this class. Initially, I also created docstrings for each method within this class docstring. However, when I rendered the documentation locally, I noticed duplicate entries, as the docstring for each method is provided already in the rest code the file. Therefore, here, I’ve only included descriptions for the `Parameters` and `Aattributes`

bobleesj: In this PR, none was modified except the class name.

bobleesj:
@sbillinge Ready for review, a few comments added via inline. Thank you.

src/diffpy/utils/scattering_objects/diffraction_object.py
bobleesj, line 1: File name has been changed from `diffraction_objects` to `diffraction_object` which carries `DiffractionObject`

sbillinge: why is the filename changed?  This would lead to an API change if we add an additional diffraction object in the future.

src/diffpy/utils/scattering_objects/diffraction_object.py
```diff
@@ -0,0 +1,492 @@
...
+    A class to manipulate diffraction data, supporting basic arithmetic operations and
+    conversions between different diffraction data metrics such as Q, two-theta, and d-spacing.
```
sbillinge, line 23: change `metrics` here to `independent variables`?

src/diffpy/utils/scattering_objects/diffraction_object.py
```diff
@@ -0,0 +1,492 @@
...
+
+    Parameters
+    ----------
+    name : str, optional
```
sbillinge, line 27: What is the difference between a parameter and an attribute here?

Also, per numpy standards, desriptions always start with "The" so wavelength would be "The wavelength of the radiation used in the diffraction experiment. Defaults to None." which, slightly magically, makes everything much more readable.  Please change everywhere.

src/diffpy/utils/scattering_objects/diffraction_object.py
```diff
@@ -0,0 +1,492 @@
...
+    Attributes
+    ----------
+    name : str
+        Name or identifier for the diffraction experiment.
+    wavelength : float
+        Wavelength of the incoming x-rays.
+    scat_quantity : str
+        Description of the type of scattering data stored.
+    on_q : list of numpy.ndarray
+        A list containing two numpy arrays: [Q values, Intensities].
```
sbillinge, line 41: The "the" constructions helps with these docstrings here, because we need a short statement of what this is, which is missing.  In this case it could be "The diffraction intensities on a grid of Q" followed by something along the ines you already put (i.e., high level view followed by pertinent details).

src/diffpy/utils/scattering_objects/diffraction_object.py
```diff
@@ -0,0 +1,492 @@
...
+
+    def set_angles_from_list(self, angles_list):
+        self.angles = angles_list
+        self.n_steps = len(angles_list) - 1.0
+        self.begin_angle = self.angles[0]
+        self.end_angle = self.angles[-1]
+
+    def set_qs_from_range(self, begin_q, end_q, step_size=None, n_steps=None):
+        """
+        create an array of linear spaced Q-values
```
sbillinge, line 202: linearly

src/diffpy/utils/scattering_objects/diffraction_object.py
```diff
@@ -0,0 +1,492 @@
...
+
+        Parameters
+        ----------
+        begin_q float
+          the beginning angle
+        end_q float
```
sbillinge, line 208: I think we usually have colons here `end_q: float`

sbillinge (commented):
few comments.  Also, please check if there were tests for `Diffraction_object`.  If there were, we should leave them until `Diffraction_object` is removed.

bobleesj:
> few comments. Also, please check if there were tests for `Diffraction_object`. If there were, we should leave them until `Diffraction_object` is removed.

Thanks for the feedback I will address each in the afternoon.

src/diffpy/utils/scattering_objects/diffraction_object.py
sbillinge: doesn't this just add useless typing though?
```
from diffpy.utils.scattering_objects.diffraction_object import DiffractionObject
from diffpy.utils.scattering_objects.pdf_object import PdfObject 
...
```
The point of the current structure is that all diffraction objects go in the diffraction_objects module.  What is the advantage of putting one per module?

src/diffpy/utils/scattering_objects/diffraction_object.py
sbillinge: I see.  So this is a direct copy-paste to duplicate it and then change the name, so we have both and they are identical to each other?

sbillinge (commented):
pls see discussion comments.

src/diffpy/utils/scattering_objects/diffraction_object.py
bobleesj: For the useless typing, if we could utilize relative import for each class. Would this be a good solution below?

```
from diffpy.utils.scattering import DiffractionObject
from diffpy.utils.scattering import PdfObject
```

bobleesj: also not sure if we want to call it as "Object" since we are utilizing a class and object is something we make with the class.. 

```
from diffpy.utils.scattering import Diffraction
from diffpy.utils.scattering import Pdf
```

This a bit more feels like pythonic?

src/diffpy/utils/scattering_objects/diffraction_object.py
bobleesj: Yes, they are identical, direct copy.

src/diffpy/utils/scattering_objects/diffraction_object.py
bobleesj: Fixed - added "The" in the front. Ref: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

bobleesj (commented):
@sbillinge ready for review again. Thank you.

src/diffpy/utils/scattering_objects/diffraction_object.py
sbillinge: This doesn't quite convey what it does though.  Diffraction is a method, a DiffractionObject is the result of a diffraction experiment.  We could call it `DiffractionDataset` or something.  I kind of like `DiffractionObject` though.   

I thought the task was to bring the current naming more into PEP8 standard, not to do a refactor?

src/diffpy/utils/scattering_objects/diffraction_object.py
sbillinge: In that case maybe let's keep it that way and not make any changes, just the name change?

sbillinge (commented):
Please see comments.  I think we are trying to do a bit too much of a refactor here.  I suggest we just do the name-change and the deprecation warnings.

If we want to remove extra typing we could simply put `DiffractionObject` into `scattering_objects.py` module.  This might make more sense.

In the future I expect we might have a PdfObject, but also others in principle, like XanesObject, RamanObject, and so on.

bobleesj:
> Please see comments. I think we are trying to do a bit too much of a refactor here. I suggest we just do the name-change and the deprecation warnings.
> 
> If we want to remove extra typing we could simply put `DiffractionObject` into `scattering_objects.py` module. This might make more sense.
> 
> In the future I expect we might have a PdfObject, but also others in principle, like XanesObject, RamanObject, and so on.

Yes, too much refactoring I agree (good discussions still It think). I created a new PR below, just focused on changing the name and deprecating it.

https://github.com/diffpy/diffpy.utils/pull/172

PR #178, opened 2024-11-15
q_to_tth & tth_to_q

### LINKED ISSUE #174: differaction objects improve functions q_to_tth, tth_to_q
### Problem
Functions do not handle invalid cases (when tth > 180 or arcsin value > 1).

### Proposed solution
Raise valueerror for those cases. Refer to #154 for more details.

yucongalicechen:
closes #174 
@sbillinge ready for review!

### DIFF
--- a/doc/source/examples/diffractionobjectsexample.rst
+++ b/doc/source/examples/diffractionobjectsexample.rst
@@ -8,34 +8,27 @@ Diffraction Objects Example
 This example will demonstrate how to use the ``DiffractionObject`` class in the
 ``diffpy.utils.scattering_objects.diffraction_objects`` module to process and analyze diffraction data.
 
-1) We have the function ``q_to_tth`` to convert q to two theta values in degrees, and ``tth_to_q`` to do the reverse.
-   You can use these functions with a pre-defined ``DiffractionObject``. ::
+1) Assuming we have created a ``DiffractionObject`` called my_diffraction_pattern from a measured diffraction pattern,
+   and we have specified the wavelength (see Section ??, to be added),
+   we can use the ``q_to_tth`` and ``tth_to_q`` functions to convert between q and two-theta. ::
 
-    # convert q to tth
-    from diffpy.utils.scattering_objects.diffraction_objects import DiffractionObject
-    test = DiffractionObject(wavelength=1.54)
-    test.on_q = [[0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]]
-    test.q_to_tth()
+    # Example: convert q to tth
+    my_diffraction_pattern.on_q = [[0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]]
+    my_diffraction_pattern.q_to_tth()
 
    This function will convert your provided q array and return a two theta array in degrees.
-   To load the converted array, you can either call ``test.q_to_tth()`` or ``test.on_q[0]``.
+   To load the converted array, you can either call ``test.q_to_tth()`` or ``test.on_q[0]``. ::
 
-   Similarly, use the function ``tth_to_q`` to convert two theta values in degrees to q values. ::
-
-    # convert tth to q
+    # Example: convert tth to q
     from diffpy.utils.scattering_objects.diffraction_objects import DiffractionObject
-    test = DiffractionObject(wavelength=1.54)
-    test.on_tth = [[0, 30, 60, 90, 120, 180], [1, 2, 3, 4, 5, 6]]
-    test.tth_to_q()
+    my_diffraction_pattern.on_tth = [[0, 30, 60, 90, 120, 180], [1, 2, 3, 4, 5, 6]]
+    my_diffraction_pattern.tth_to_q()
 
-   To load the converted array, you can either call ``test.tth_to_q()`` or ``test.on_tth[0]``.
+   Similarly, to load the converted array, you can either call ``test.tth_to_q()`` or ``test.on_tth[0]``.
 
-2) You can use these functions without specifying a wavelength. However, if so, the function will return an empty array,
-   so we strongly encourage you to specify a wavelength when using these functions. ::
+2) Both functions require a wavelength to perform conversions. Without a wavelength, they will return empty arrays.
+   Therefore, we strongly encourage you to specify a wavelength when using these functions. ::
 
-    from diffpy.utils.scattering_objects.diffraction_objects import DiffractionObject
-    test = DiffractionObject()
-    test.on_q = [[0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]]
-    test.q_to_tth()
-
-   In this case, the function will return an empty array on two theta.
+    # Example: without wavelength specified
+    my_diffraction_pattern.on_q = [[0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]]
+    my_diffraction_pattern.q_to_tth()    # returns an empty array
--- a/doc/source/examples/diffractionobjectsexample.rst
+++ b/doc/source/examples/diffractionobjectsexample.rst
@@ -1,0 +1,41 @@
+.. _Diffraction Objects Example:
+
+:tocdepth: -1
+
+Diffraction Objects Example
+###########################
+
+This example will demonstrate how to use the ``DiffractionObject`` class in the
+``diffpy.utils.scattering_objects.diffraction_objects`` module to process and analyze diffraction data.
+
+1) We have the function ``q_to_tth`` to convert q to two theta values in degrees, and ``tth_to_q`` to do the reverse.
+   You can use these functions with a pre-defined ``DiffractionObject``. ::
+
+    # convert q to tth
+    from diffpy.utils.scattering_objects.diffraction_objects import DiffractionObject
+    test = DiffractionObject(wavelength=1.54)
+    test.on_q = [[0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]]
+    test.q_to_tth()
+
+   This function will convert your provided q array and return a two theta array in degrees.
+   To load the converted array, you can either call ``test.q_to_tth()`` or ``test.on_q[0]``.
+
+   Similarly, use the function ``tth_to_q`` to convert two theta values in degrees to q values. ::
+
+    # convert tth to q
+    from diffpy.utils.scattering_objects.diffraction_objects import DiffractionObject
+    test = DiffractionObject(wavelength=1.54)
+    test.on_tth = [[0, 30, 60, 90, 120, 180], [1, 2, 3, 4, 5, 6]]
+    test.tth_to_q()
+
+   To load the converted array, you can either call ``test.tth_to_q()`` or ``test.on_tth[0]``.
+
+2) You can use these functions without specifying a wavelength. However, if so, the function will return an empty array,
+   so we strongly encourage you to specify a wavelength when using these functions. ::
+
+    from diffpy.utils.scattering_objects.diffraction_objects import DiffractionObject
+    test = DiffractionObject()
+    test.on_q = [[0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]]
+    test.q_to_tth()
+
+   In this case, the function will return an empty array on two theta.
--- a/doc/source/utilities/diffractionobjectsutility.rst
+++ b/doc/source/utilities/diffractionobjectsutility.rst
@@ -1,0 +1,16 @@
+.. _Diffraction Objects Utility:
+
+Diffraction Objects Utility
+===========================
+
+The ``diffpy.utils.scattering_objects.diffraction_objects`` module provides functions
+for managing and analyzing diffraction data, including angle-space conversions
+and interactions between diffraction data.
+
+- ``q_to_tth()``: Converts an array of q values to their corresponding two theta values, based on specified wavelength.
+- ``tth_to_q()``: Converts an array of two theta values to their corresponding q values, based on specified wavelength.
+
+  These functions help developers standardize diffraction data and update the arrays
+  in the associated ``DiffractionObject``, enabling easier analysis and further processing.
+
+For a more in-depth tutorial for how to use these functions, click :ref:`here <Diffraction Objects Example>`.
--- a/doc/source/utilities/diffractionobjectsutility.rst
+++ b/doc/source/utilities/diffractionobjectsutility.rst
@@ -7,10 +7,7 @@ The ``diffpy.utils.scattering_objects.diffraction_objects`` module provides func
 for managing and analyzing diffraction data, including angle-space conversions
 and interactions between diffraction data.
 
-- ``q_to_tth()``: Converts an array of q values to their corresponding two theta values, based on specified wavelength.
-- ``tth_to_q()``: Converts an array of two theta values to their corresponding q values, based on specified wavelength.
-
-  These functions help developers standardize diffraction data and update the arrays
-  in the associated ``DiffractionObject``, enabling easier analysis and further processing.
+These functions help developers standardize diffraction data and update the arrays
+in the associated ``DiffractionObject``, enabling easier analysis and further processing.
 
 For a more in-depth tutorial for how to use these functions, click :ref:`here <Diffraction Objects Example>`.
--- a/src/diffpy/utils/scattering_objects/diffraction_objects.py
+++ b/src/diffpy/utils/scattering_objects/diffraction_objects.py
@@ -763,25 +763,29 @@ class DiffractionObject:
 
             2\theta_n = 2 \arcsin\left(\frac{\lambda q}{4 \pi}\right)
 
+        Function adapted from scikit-beam.  Thanks to those developers
+
         Parameters
         ----------
         q : array
-            An array of :math:`q` values
+            The array of :math:`q` values
 
         wavelength : float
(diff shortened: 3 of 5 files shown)


src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -763,25 +763,29 @@ def q_to_tth(self):
 
             2\theta_n = 2 \arcsin\left(\frac{\lambda q}{4 \pi}\right)
 
+        Function adapted from scikit-beam.  Thanks to those developers
```
yucongalicechen, line 766: Pull the line here because I think it's causing issues in the API... which takes it as a parameter.

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -800,25 +804,25 @@ def tth_to_q(self):
...
-
         Returns
         -------
         q : array
-            An array of :math:`q` values in the inverse of the units
+            The array of :math:`q` values in the inverse of the units
             of ``wavelength``
         """
         two_theta = np.asarray(np.deg2rad(self.on_tth[0]))
+        if np.any(two_theta > np.pi):
```
sbillinge, line 824: as above, I think this isn't doing quite what we want.

sbillinge (commented):
pls see comments inline.  Per our earlier convo, I think we should be working on these in the tests first.  Let's converge the tests first, then write the code.  What is the behavior that we want to handle?  I know it is something like tth exceeding 180 deg, but when would this come about in practice?  We only need to handle it if it is a user-facing issue, so what would a user do where this would be encountered?  then how do we want the code to behave to alert the user?

src/diffpy/utils/scattering_objects/diffraction_objects.py
```diff
@@ -763,25 +763,27 @@ def q_to_tth(self):
...
         two_theta : array
-            An array of :math:`2\theta` values in radians
+            The array of :math:`2\theta` values in radians
         """
         q = self.on_q[0]
         q = np.asarray(q)
         wavelength = float(self.wavelength)
         pre_factor = wavelength / (4 * np.pi)
+        if np.any(np.abs(q * pre_factor) > 1):
+            raise ValueError("Please check if you entered an incorrect wavelength or q value.")
```
yucongalicechen, line 786: I think we would want to raise a value error here because otherwise the code proceeds with tth = nan.

sbillinge (commented):
Let's first work on the test and then the code, so don't modify the code yet, just the test. The test should test the error message that is sent to users under different situations.  The most likely one we discussed is that they give a wrong wavelength.  They could also give no wavelength. What behavior do we want then?  Less likely but possible is they give a tth array that goes above 180. What behavior do we want then?

yucongalicechen:
> Let's first work on the test and then the code, so don't modify the code yet, just the test. The test should test the error message that is sent to users under different situations. The most likely one we discussed is that they give a wrong wavelength. They could also give no wavelength. What behavior do we want then? Less likely but possible is they give a tth array that goes above 180. What behavior do we want then?

Yes, sorry, I added more tests now. @sbillinge it looks like in DO tth is implemented on degree scale only, shall we make another issue to allow tth to compute in radian?

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -232,25 +232,52 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
     expected_tth = [0, 23.07392, 47.15636, 73.73980, 106.26020, 180]
     assert np.allclose(actual_tth, expected_tth)
 
 
-def test_q_to_tth_bad():
-    # invalid wavelength or q values when arcsin value is not in the range of [-1, 1]
-    actual = DiffractionObject(wavelength=4 * np.pi)
-    setattr(actual, "on_q", [[0.6, 0.8, 1, 1.2], [1, 2, 3, 4]])
+params_q_to_tth_bad = [
+    # UC1: user did not specify wavelength
```
sbillinge, line 245: Do we want to not allow missing wavelengths? This makes these DOs harder to use, but also makes them less useful. There may be a middle ground where we allow it, but let the user know that much of the functionality goes away without it.  We could also trigger a workflow that requests the wavelength, but still allows users to override that. This would encourage them to enter a wavelength but not insist?

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -232,25 +232,52 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+    (
+        [None, [0, 0.2, 0.4, 0.6, 0.8, 1]],
+        "Wavelength is not specified. Please provide a valid wavelength, "
+        "e.g., DiffractionObject(wavelength=0.71).",
+    ),
+    # UC2: user specified invalid q values that result in tth > 180 degrees
+    (
+        [4 * np.pi, [0.2, 0.4, 0.6, 0.8, 1, 1.2]],
+        "Wavelength * q > 4 * pi. Please check if you entered an incorrect wavelength or q value.",
```
sbillinge, line 254: I think this behavior is good.  The first sentence is a bit mathematical and cryptic. Is there a more "chemist friendly" way of saying it?

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -232,25 +232,52 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+    ),
+    # UC3: user specified a wrong wavelength that result in tth > 180 degrees
+    (
+        [100, [0, 0.2, 0.4, 0.6, 0.8, 1]],
+        "Wavelength * q > 4 * pi. Please check if you entered an incorrect wavelength or q value.",
+    ),
+    # UC4: user specified an empty q array
```
sbillinge, line 261: It is ok to specify no q-array (you can instantiate on tth for example).  Presumably this is only an error if the the DO is being created with q data.  So the behavior is ok, but we may want to tweak the error message.  I think a more general error is if the x and y arrays are not the same length, whether they are q, tth or d.  I suggest to handle that more general case?

Finally, do we also want to specify what error is raised as well as the message?

sbillinge (commented):
Please see inline

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
yucongalicechen: Yes this sounds good. I think user can also directly set attributes so that they can use DO without a wavelength. I would suggest to prompt user inputs in the `insert_scattering_quantity` function instead of this one when it wants to set arrays on all tth/q/dspace, so that it can avoid calling these functions?
When people use this function directly we would want them to speicify a wavelength, so it's better if we raise an error?

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
yucongalicechen: Does "The combination of wavelength and q values is too large" sound good?

sbillinge: How about: "the supplied q-array and wavelength will result in an impossible two-theta.  Please check these values and re-instantiate the DiffractionObject"

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
sbillinge: For now, how about we just print a warning message if no wavelength is supplied. Something like "INFO: no wavelength has been specified. You can continue to use the DiffractionObject but some of it's powerful features will not be available. To specify a wavelength....<instructions>"

yucongalicechen:
@sbillinge ready for review

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -231,94 +231,179 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+    when there is no wavelength, we expect the correct warning message and output,
+    otherwise, we only check the output matches the expected array."""
+    if actual_diffraction_object.wavelength is None:
+        with pytest.warns(UserWarning) as warn_record:
+            getattr(actual_diffraction_object, function)()
+        assert str(warn_record[0].message) == (
+            "INFO: no wavelength has been specified. You can continue "
+            "to use the DiffractionObject but some of its powerful features "
+            "will not be available. To specify a wavelength, you can use "
+            "DiffractionObject(wavelength=0.71)."
```
sbillinge, line 245: 0.71?  Or something like "wavelength=<wavelength>) where you replace <wavelength> with the actual wavelength in the units of..." Etc.   Also, are these the correct instructions? This seems to be instantiating a new DO not adding a wavelength to an existing DO.

Let's also make sure to add these instructions to the documentation.

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -231,94 +231,179 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+    # UC1: user specified invalid q values that result in tth > 180 degrees
     (
-        [None, [0, 0.2, 0.4, 0.6, 0.8, 1]],
-        "Wavelength is not specified. Please provide a valid wavelength, "
-        "e.g., DiffractionObject(wavelength=0.71).",
+        [4 * np.pi, [0.2, 0.4, 0.6, 0.8, 1, 1.2], [1, 2, 3, 4, 5, 6]],
+        [
+            ValueError,
+            "The supplied q-array and wavelength will result in an impossible two-theta. "
+            "Please check these values and re-instantiate the DiffractionObject.",
```
sbillinge, line 292: Maybe add ."... With correct values"

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -231,94 +231,179 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
     ),
-    # UC2: user specified invalid q values that result in tth > 180 degrees
+    # UC2: user specified a wrong wavelength that result in tth > 180 degrees
     (
-        [4 * np.pi, [0.2, 0.4, 0.6, 0.8, 1, 1.2]],
-        "Wavelength * q > 4 * pi. Please check if you entered an incorrect wavelength or q value.",
+        [100, [0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]],
+        [
+            ValueError,
+            "The supplied q-array and wavelength will result in an impossible two-theta. "
```
sbillinge, line 300: See above.  Also since we are reusing the error message do we want to minimize word by defining it in a variable once and reusing the variable in multiple tests?

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -231,94 +231,179 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
     ),
-    # UC4: user specified an empty q array
-    ([4 * np.pi, []], "Q array is empty. Please provide valid q values."),
-    # UC5: user specified a non-numeric value in q array
+    # UC4: user specified a q array that does not match the length of intensity array (with wavelength)
     (
-        [4 * np.pi, [0, 0.2, 0.4, 0.6, 0.8, "invalid"]],
-        "Invalid value found in q array. Please ensure all values are numeric.",
+        [4 * np.pi, [0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5]],
+        [IndexError, "Please ensure q array and intensity array are the same length."],
```
sbillinge, line 312: I think RunTime wrote may be more appropriate here since I think we will raise it ourselves?

sbillinge (commented):
Please see inline

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
yucongalicechen: Yes I will store them in diffraction_objects.py file

yucongalicechen:
@sbillinge thanks for the comments. Since the tests are almost there I'll start implementing the functions.

tests/diffpy/utils/scattering_objects/test_diffraction_objects.py
```diff
@@ -231,6 +231,158 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
     assert (diffraction_object1 == diffraction_object2) == expected
 
 
+def _test_valid_diffraction_objects(actual_diffraction_object, function, expected_array):
+    """Checks the behavior of the DiffractionObject:
```
sbillinge, line 235: Test functions don't need docstring, though if it's ok to leave this now you did it.

sbillinge (commented):
This looks good now good to go on the code.

sbillinge:
> > looks good. this needs a news and possibly some update in the docs. You don't have to explain error messages in docs, but it might be nice to explain correct usage.
> 
> @sbillinge I add the news. I can add another issue on adding example/utility for functionalities we updated this PR?

Why not just do it on this PR?

doc/source/examples/diffractionobjectsexample.rst
```diff
@@ -0,0 +1,41 @@
...
+
+This example will demonstrate how to use the ``DiffractionObject`` class in the
+``diffpy.utils.scattering_objects.diffraction_objects`` module to process and analyze diffraction data.
+
+1) We have the function ``q_to_tth`` to convert q to two theta values in degrees, and ``tth_to_q`` to do the reverse.
+   You can use these functions with a pre-defined ``DiffractionObject``. ::
+
+    # convert q to tth
+    from diffpy.utils.scattering_objects.diffraction_objects import DiffractionObject
+    test = DiffractionObject(wavelength=1.54)
```
sbillinge, line 16: instead of test let's call it `my_diffraction_pattern`

We are not writing all of the docs here, so you can start with something  like "assuming we have created a `DiffractionObject` called `my_diffraction_pattern` from a measured diffraction pattern, and we have specified the wavelenth (see Section ??)  we can use the `q_to_tth` and `tth_to_q` functions to convert between Q and two-theta by typing `my_diffraction_pattern.q_to_tth()`..." ..and so on.   Let's make an issue to add to docs "How to set a wavelength" and how to instantiate a diffraction object and how to add a diffraction pattern.

doc/source/utilities/diffractionobjectsutility.rst
```diff
@@ -0,0 +1,16 @@
+.. _Diffraction Objects Utility:
+
+Diffraction Objects Utility
+===========================
+
+The ``diffpy.utils.scattering_objects.diffraction_objects`` module provides functions
+for managing and analyzing diffraction data, including angle-space conversions
+and interactions between diffraction data.
+
+- ``q_to_tth()``: Converts an array of q values to their corresponding two theta values, based on specified wavelength.
```
sbillinge, line 10: we don't need this here.  It will be auto-generated in the API docs, and it is shown as an aexample above.

sbillinge (commented):
nicely done there!  Please see my comments and then I can merge this.

yucongalicechen:
@sbillinge ready for review! I added an issue for additional docs (#183).

sbillinge (commented):
I think we don't need these methods at all, at least as public methods.  Please see my comment.

doc/source/examples/diffractionobjectsexample.rst
```diff
@@ -8,34 +8,27 @@ Diffraction Objects Example
...
-    test.q_to_tth()
+    # Example: convert q to tth
+    my_diffraction_pattern.on_q = [[0, 0.2, 0.4, 0.6, 0.8, 1], [1, 2, 3, 4, 5, 6]]
+    my_diffraction_pattern.q_to_tth()
 
    This function will convert your provided q array and return a two theta array in degrees.
-   To load the converted array, you can either call ``test.q_to_tth()`` or ``test.on_q[0]``.
-
-   Similarly, use the function ``tth_to_q`` to convert two theta values in degrees to q values. ::
+   To load the converted array, you can either call ``test.q_to_tth()`` or ``test.on_q[0]``. ::
```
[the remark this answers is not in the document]
yucongalicechen: I think this is a private function for now (I didn't see the "_" in the function name)? We need to call `insert_scattering_quantity` in order for it to populate on all arrays automatically. If they just do `my_diffraction_pattern = DiffractionObjects()` and `my_diffraction_pattern.on_q = ...` it is not automatically populated to tth.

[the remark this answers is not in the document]
sbillinge: but if they do that they are just instantiating an empty DO so there is nothing to populate anywhere..... So that is desired behavior.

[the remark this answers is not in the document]
sbillinge: The question is what behavior do we want?  I want to be able to get my intensity data on all the different x-grids, but I can do that by typing `my_pattern.on_q` orwhatever.  What is the UC where I would want to run that function?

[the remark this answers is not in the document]
yucongalicechen: I see. Does this sound better?

Assuming we have created a ``DiffractionObject`` called my_diffraction_pattern from a measured diffraction pattern, and we have specified the wavelength (see Section ??, to be added), we can use the ``q_to_tth`` and ``tth_to_q`` functions to convert between q and two-theta. For example, ``my_diffraction_pattern.q_to_tth()`` converts q to two-theta, while ``my_diffraction_pattern.tth_to_q()`` converts two-theta to q. The converted array can be accessed using by calling ``my_diffraction_pattern.on_q[0]`` for q, or ``my_diffraction_pattern.on_tth[0]`` for two-theta.

PR #184, opened 2024-12-01
refactor the transforms out of DiffractionObjects

sbillinge:
to make them more widely available in general

### DIFF
--- a/src/diffpy/utils/transforms.py
+++ b/src/diffpy/utils/transforms.py
@@ -1,0 +1,122 @@
+import warnings
+from copy import copy
+
+import numpy as np
+
+wavelength_warning_emsg = (
+    "INFO: no wavelength has been specified. You can continue "
+    "to use the DiffractionObject but some of its powerful features "
+    "will not be available. To specify a wavelength, set "
+    "diffraction_object.wavelength = [number], "
+    "where diffraction_object is the variable name of you Diffraction Object, "
+    "and number is the wavelength in angstroms."
+)
+invalid_tth_emsg = "Two theta exceeds 180 degrees. Please check the input values for errors."
+invalid_q_or_wavelength_emsg = (
+    "The supplied q-array and wavelength will result in an impossible two-theta. "
+    "Please check these values and re-instantiate the DiffractionObject with correct values."
+)
+
+
+def _validate_inputs(on_q, wavelength):
+    if wavelength is None:
+        warnings.warn(wavelength_warning_emsg, UserWarning)
+        return np.empty(0)
+    pre_factor = wavelength / (4 * np.pi)
+    if np.any(np.abs(on_q[0] * pre_factor) > 1.0):
+        raise ValueError(invalid_q_or_wavelength_emsg)
+
+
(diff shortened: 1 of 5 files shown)


src/diffpy/utils/transforms.py
```diff
@@ -0,0 +1,122 @@
...
+def _validate_inputs(on_q, wavelength):
+    if wavelength is None:
+        warnings.warn(wavelength_warning_emsg, UserWarning)
+        return np.empty(0)
+    pre_factor = wavelength / (4 * np.pi)
+    if np.any(np.abs(on_q[0] * pre_factor) > 1.0):
+        raise ValueError(invalid_q_or_wavelength_emsg)
+
+
+def q_to_tth(on_q, wavelength):
```
8bitsam, line 30: I wonder if it might be better to have the function parameter be a `Diffraction_object`, and extract q and two theta from that object inside the body of the function. This would keep it consistent to the way it was before in that you can avoid having to reference instance variables in a function call. I might be wrong about this, but I don't think it's likely the functions would be used without diffraction objects, so this would be a bit of a "shorthand" compared to what it is now.

src/diffpy/utils/transforms.py
```diff
@@ -0,0 +1,122 @@
...
+    on_tth = copy(on_q)  # initialize output array of same shape
+    if wavelength is not None:
+        on_tth[0] = np.rad2deg(2.0 * np.arcsin(on_q[0] * wavelength / (4 * np.pi)))
+    else:  # return intensities vs. an x-array that is just the index
+        for i, _ in enumerate(on_q[0]):
+            on_tth[0][i] = i
+    return on_tth
+
+
+def tth_to_q(on_tth, wavelength):
```
8bitsam, line 76: Same comment as `q_to_tth`.

src/diffpy/utils/transforms.py
sbillinge: Thanks for the suggestion @8bitsam .  I chose not to do that because then you would need to instantiate a DO before you could do a transform on your data.  The point of pulling it out of the DO and into `transforms` was so that people who weren't using DOs could still use the transforms.  Perhaps it would even make sense to not require a 2D np array which is uncommon for people to have in general (they often hand around 1D vector arrays).  It does seem odd to have to break up our DO and then reformulate it, which is what results from this.   I think this is not a problem if there is not performance issue, because the DO is using it as a private function.  My initial intention was to make it a private function because no user ever using DOs will need to call it.

yucongalicechen: I was also thinking that maybe we shall pass in the x-array only, since we won't change anything to the intensity array in the function. It does seem to match with the DO better so maybe we should keep the 2D array.

src/diffpy/utils/transforms.py
```diff
@@ -0,0 +1,122 @@
...
+    r"""
+    Helper function to convert two-theta to q on independent variable axis.
```
yucongalicechen, line 78: It looks like we only have our tth in degrees during the conversion while radian is also an option for x-array. We should specify this somewhere around here?

PR #190, opened 2024-12-03
move tests to top level directory

sbillinge:


sbillinge:
@bobleesj @yucongalicechen @Sparks29032 I am not sure why the tests before were all arranged in different sub-directories so I consolidated them, but if this is going to cause issues we can not do this.  Please let me know if there was a reason for that.

Issue #186, opened 2024-12-02
write tests for


sbillinge:
### Problem
- [x] d_to_tth
- [x] tth_to_d
- [x] d_to_q
- [x] q_to_d
- [x] get_angle_index
- [ ] scale_to
- [x] on x_type

PR #191, opened 2024-12-03
test function for get_angle_index

yucongalicechen:
issue #186: test function for `get_angle_index`

@sbillinge ready for review


tests/test_diffraction_objects.py
```diff
@@ -241,6 +241,24 @@ def _test_valid_diffraction_objects(actual_diffraction_object, function, expecte
     return np.allclose(actual_array, expected_array)
 
 
+def test_get_angle_index():
+    test = DiffractionObject()
+    test.angles = np.array([10, 20, 30, 40, 50, 60])
```
sbillinge, line 246: I guess we would want to look in `on_tth[0]` not self. Angles.  I'm fact do we need self.angles?

sbillinge (commented):
We may want to generalize this to get the index of any x-array, not just angles?  Also, I could think of more edge cases so we probably want to use purest.parametrize

tests/test_diffraction_objects.py
yucongalicechen: Yes it looks like angles should be replaced by on_tth[0]

yucongalicechen: Shall we do get_array_index then? User specifies the xtype and target value, and we return the index they're looking for

sbillinge: > Shall we do get_array_index then? User specifies the xtype and target value, and we return the index they're looking for

Yes, that sounds about right

src/diffpy/utils/diffraction_objects.py
```diff
@@ -248,15 +248,27 @@ def _set_array_from_range(self, begin, end, step_size=None, n_steps=None):
             array = np.linspace(begin, end, n_steps)
         return array
 
-    def get_angle_index(self, angle):
-        count = 0
-        for i, target in enumerate(self.angles):
-            if angle == target:
+    def get_array_index(self, xtype, value):
```
yucongalicechen, line 251: I am wondering if it is better to set default xtype to self.xtype, so that user only needs to enter the value when searching for an index. The xtype parameter would only need to be specified if the user wants to find the index of a different xarray.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -248,15 +248,27 @@ def _set_array_from_range(self, begin, end, step_size=None, n_steps=None):
...
+            the xtype used to access the array
+        value float
+            the target value to search for
+
+        Returns
+        -------
+        the index of the value in the array
+        """
+        if self.on_xtype(xtype) is None:
+            raise ValueError(_xtype_wmsg(xtype))
```
yucongalicechen, line 267: I think I can remove this if I let `on_xtype()` to raise an error for invalid xtypes. Will do that in the other PR.

yucongalicechen:
@sbillinge ready for review!

src/diffpy/utils/diffraction_objects.py
sbillinge: yes I like that.

tests/test_diffraction_objects.py
```diff
@@ -212,21 +213,41 @@ def _test_valid_diffraction_objects(actual_diffraction_object, function, expecte
 
 
 def test_get_angle_index():
```
sbillinge, line 215: I think we need to rename this.

tests/test_diffraction_objects.py
```diff
@@ -212,21 +213,41 @@ def _test_valid_diffraction_objects(actual_diffraction_object, function, expecte
...
+    )
+    actual_index = test.get_array_index(xtype="tth", value=30)
+    assert actual_index == 0
+
+
+params_index_bad = [
+    # UC1: empty array
+    (
+        [0.71, np.array([]), np.array([]), "tth", "tth", 10],
+        [IndexError, "WARNING: no matching value 10 found in the tth array."],
```
sbillinge, line 227: is this an error or a warning?

tests/test_diffraction_objects.py
```diff
@@ -212,21 +213,41 @@ def _test_valid_diffraction_objects(actual_diffraction_object, function, expecte
...
 
 def test_get_angle_index():
-    test = DiffractionObject()
-    test.angles = np.array([10, 20, 30, 40, 50, 60])
-    actual_angle_index = test.get_angle_index(angle=10)
-    assert actual_angle_index == 0
+    test = DiffractionObject(
+        wavelength=0.71, xarray=np.array([30, 60, 90]), yarray=np.array([1, 2, 3]), xtype="tth"
+    )
+    actual_index = test.get_array_index(xtype="tth", value=30)
```
sbillinge, line 219: we need more cases.  Here everything is integers and there is a match.  What do we want to happen if the value lies between two other values?  Return nearest and a warning?  What if it lies outside the range of values?  Return nearest and a warning?  What if it is really far away?  Have a threshold after which we raise and error?

sbillinge (commented):
please see inline

src/diffpy/utils/diffraction_objects.py
```diff
@@ -263,12 +263,30 @@ def get_array_index(self, xtype, value):
...
-        for i, target in enumerate(self.on_xtype(xtype)[0]):
-            if value == target:
-                return i
-        raise IndexError(f"WARNING: no matching value {value} found in the {xtype} array.")
+
+        if xtype is None:
+            xtype = self.input_xtype
+        if self.on_xtype(xtype) is None or len(self.on_xtype(xtype)[0]) == 0:
+            raise ValueError(
+                f"The '{xtype}' array is empty. " "Please ensure it is initialized and the correct xtype is used."
```
yucongalicechen, line 271: The current `on_xtype` function does not raise an error when the specified xtype is invalid. I was thinking about letting that function raise an error in those cases, but it looks like we can handle invalid xtype together with empty array.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -263,12 +263,30 @@ def get_array_index(self, xtype, value):
...
+            )
+        array = self.on_xtype(xtype)[0]
+        i = (np.abs(array - value)).argmin()
+        nearest_value = np.abs(array[i] - value)
+        distance = min(np.abs(value - array.min()), np.abs(value - array.max()))
+        threshold = 0.5 * (array.max() - array.min())
```
yucongalicechen, line 277: I am not super sure about the threshold here. I'm thinking about a dynamic threshold within the reasonable range of each array, but this here can cause a bit of problem if the array is sparse.

tests/test_diffraction_objects.py
yucongalicechen: Should be an error. I've edited the error message.

tests/test_diffraction_objects.py
yucongalicechen: @sbillinge I addressed this in the new PR. please review.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -248,15 +248,45 @@ def _set_array_from_range(self, begin, end, step_size=None, n_steps=None):
...
+
+        if nearest_value != 0 and (array.min() <= value <= array.max() or distance <= threshold):
```
sbillinge, line 279: I am rethinking all this.  I wonder if we should just return i and be done with it.  Raise an error if the array is empty as you do, but then just return i.  what do you think?  Then it is up to the user to make sure the result makes sense.

sbillinge (commented):
please see inline

sbillinge:
this needs a news too.

PR #192, opened 2024-12-03
test function on_xtype

yucongalicechen:
issue #186: test function for `on_xtype`


sbillinge:
I am a little unclear what this is doing.  Could you maybe hit down some words to explain the high level view?

yucongalicechen:
> I am a little unclear what this is doing. Could you maybe hit down some words to explain the high level view?

I think this is for checking if `on_xtype` function returns the correct 2D array (e.g., tth array, q array, etc.)

sbillinge:
> > I am a little unclear what this is doing. Could you maybe hit down some words to explain the high level view?
> 
> I think this is for checking if `on_xtype` function returns the correct 2D array (e.g., tth array, q array, etc.)

Is this a new function? One that can be used instead of `self.on_tth`? We want that function, but then we want a test that tests that it returns the right thing.

yucongalicechen:
> > > I am a little unclear what this is doing. Could you maybe hit down some words to explain the high level view?
> > 
> > 
> > I think this is for checking if `on_xtype` function returns the correct 2D array (e.g., tth array, q array, etc.)
> 
> Is this a new function? One that can be used instead of `self.on_tth`? We want that function, but then we want a test that tests that it returns the right thing.

It's an old function. This PR is to check that it returns the right array and raise an error if it's an invalid xtype. It doesn't replace on_tth but simply calls on_tth if xtype is tth.

tests/test_diffraction_objects.py
```diff
@@ -232,13 +238,40 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+            np.array([[10, 20, 30, 40, 50, 60], [100, 200, 300, 400, 500, 600]]),  # expected on_d
+        ],
+    )
+]
+
+
+@pytest.mark.parametrize("inputs, expected", params_on_xtype)
+def test_on_xtype(inputs, expected):
+    test = DiffractionObject()
+    test.on_tth = np.array([inputs[1], inputs[0]])
```
sbillinge, line 261: This is not really how DO's should be used.  The tth, q and d arrays are somehow all different versions of the same array so we shouldn't be able to set them independently like this.  If we can now this is not desirable behavior, so we probably don't want a test that does this.    We should probably create a `set` property operator that generates all the arrays whenever a single one is specified in this way.  But for sure, we don't want a test with this undesirable behavior....

tests/test_diffraction_objects.py
```diff
@@ -232,13 +238,40 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
     assert (diffraction_object1 == diffraction_object2) == expected
 
 
-def _test_valid_diffraction_objects(actual_diffraction_object, function, expected_array):
```
sbillinge, line 235: is there a reason this is getting deleted?

sbillinge (commented):
pls see inline comment

src/diffpy/utils/diffraction_objects.py
```diff
@@ -394,7 +395,7 @@ def on_xtype(self, xtype):
         elif xtype.lower() in DQUANTITIES:
             return self.on_d()
         else:
-            warnings.warn(_xtype_wmsg(xtype))
+            raise ValueError(_xtype_wmsg(xtype))
```
yucongalicechen, line 398: Is it better if we raise an error here?

tests/test_diffraction_objects.py
yucongalicechen: It's not used anymore. It was originally designed for testing q_to_tth and tth_to_q.

yucongalicechen:
@sbillinge ready for review!

src/diffpy/utils/diffraction_objects.py
```diff
@@ -376,25 +376,26 @@ def scale_to(self, target_diff_object, xtype=None, xvalue=None):
         return scaled
 
     def on_xtype(self, xtype):
-        """
+        f"""
         return a 2D np array with x in the first column and y in the second for x of type type
```
sbillinge, line 380: I think we want to return a list of two 1D arrays

src/diffpy/utils/diffraction_objects.py
```diff
@@ -376,25 +376,26 @@ def scale_to(self, target_diff_object, xtype=None, xvalue=None):
...
         Parameters
         ----------
-        xtype
+        xtype str
+            the type of quantity for the independent variable from {*XQUANTITIES, }
 
         Returns
         -------
-
+        a 2D np array with x and y data
```
sbillinge, line 389: see above

src/diffpy/utils/diffraction_objects.py
sbillinge: I am not sure I understand....we are raising an error...

tests/test_diffraction_objects.py
```diff
@@ -208,38 +202,22 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
-    test = DiffractionObject()
-    test.on_tth = np.array([inputs[1], inputs[0]])
-    test.on_q = np.array([inputs[2], inputs[0]])
-    test.on_d = np.array([inputs[3], inputs[0]])
-    for xtype_list, expected_value in zip([ANGLEQUANTITIES, QQUANTITIES, DQUANTITIES], expected):
-        for xtype in xtype_list:
-            assert np.allclose(test.on_xtype(xtype), expected_value)
+def test_on_xtype():
+    test = DiffractionObject(wavelength=2 * np.pi, xarray=np.array([30, 60]), yarray=np.array([1, 2]), xtype="tth")
+    assert np.allclose(test.on_xtype("tth"), [np.array([30, 60]), np.array([1, 2])])
```
sbillinge, line 207: please see above.  I think we will return a list of 1D arrays

sbillinge (commented):
please see inline..  We can discuss, but this should behave the same as `DO.on_q()` etc., and I think they should maybe return list of 1D arrays.

tests/test_diffraction_objects.py
```diff
@@ -202,13 +202,24 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
+    assert np.allclose(test.on_xtype("q"), [np.array([0.51764, 1]), np.array([1, 2])])
+    assert np.allclose(test.on_xtype("d"), [np.array([12.13818, 6.28319]), np.array([1, 2])])
+
+
+def test_on_xtype_bad():
+    test = DiffractionObject()
+    with pytest.raises(
+        ValueError,
+        match=re.escape(
+            f"WARNING: I don't know how to handle the xtype, 'invalid'.  Please rerun specifying an "
```
sbillinge, line 218: I think are now raising an exception, right? So this isn't a warning.  Just remove `WARNING:`    Python will handle the rest.

sbillinge (commented):
one small tweak inline

PR #193, opened 2024-12-04
DiffractionObject can now be instantiated directly

### LINKED ISSUE #177: diffraction objects set_all_arrays
After #174, #175, #176 are closed, call these functions in set_all_arrays.
Refer to #154 for more details.

### LINKED ISSUE #181: Allow initialization of DO on instantiation
### Problem

the workflow now is:
```
my_pattern = DiffractionObject(wavelength=1.54)
my_pattern.insert_scattering_oject(x_array, y_array, "tth", metadata={"something":"thing"})
```

### Proposed solution
As an option we would like to support this:
```
my_pattern = DiffractionObject(x=x_array, y=y_array, dtype="tth", wavelength=1.54, metadata={"something":"thing"})
```
or something like that.

### LINKED ISSUE #185: refactor base type for DiffractionObject data
### Problem

currently the data are  `np.array(2,len(intensity))`, but more standard would probably be `np.array(len(intensity),2) , and at that point we could put all the other independent variable axes in the same array to avoid duplication, so `np.array(len(q), num_axes)`, something like that.  we could use standard that intensity data is in column 0, col 1 is q, col 2 is tth, col 3 is d and so on.

### Proposed solution
see above

sbillinge:
closes #185 
closes #181
closes #177

and basic data structure is a 2D array with everything in it

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -1,11 +1,10 @@
 import datetime
-import warnings
 from copy import deepcopy
 
 import numpy as np
 
 from diffpy.utils.tools import get_package_info
-from diffpy.utils.transforms import q_to_tth, tth_to_q
+from diffpy.utils.transforms import d_to_q, d_to_tth, q_to_d, q_to_tth, tth_to_d, tth_to_q
 
 QQUANTITIES = ["q"]
 ANGLEQUANTITIES = ["angle", "tth", "twotheta", "2theta"]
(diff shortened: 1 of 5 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -19,398 +18,17 @@
...
-        self._all_arrays = [self.on_q, self.on_tth]
-        self.metadata = {}
+    def __init__(
+        self, name="", wavelength=None, scat_quantity="", metadata={}, xarray=None, yarray=None, xtype=""
+    ):
+        if xarray is None:
+            xarray = np.empty(0)
+        if yarray is None:
+            yarray = np.empty(0)
+        self.insert_scattering_quantity(
```
bobleesj, line 29: nice re-use of `insert_scattering_quantity`

src/diffpy/utils/diffraction_objects.py
```diff
@@ -652,42 +270,47 @@ def insert_scattering_quantity(
...
-        if name is not None:
-            self.name = name
-        if wavelength is not None:
-            self.wavelength = wavelength
+        self.metadata = metadata
+        self.scat_quantity = scat_quantity
+        self.name = name
+        self.wavelength = wavelength
+        self.all_arrays = np.empty(shape=(len(yarray), 4))
+        self.all_arrays[:, 0] = yarray
```
bobleesj, line 303: This is much much easier to read now!

tests/test_diffraction_objects.py
```diff
@@ -7,229 +7,193 @@
...
-            [np.array([1.00001, 2.00001]), np.array([3.00001, 4.00001])],
-            [np.array([1.00001, 2.00001]), np.array([3.00001, 4.00001])],
-            {"thing1": 1, "thing2": "thing2"},
-        ],
+        {
+            "name": "same",
+            "scat_quantity": "x-ray",
+            "wavelength": 0.71,
+            "xtype": "q",
+            "xarray": np.array([1.0, 2.0]),
```
bobleesj, line 59: Much nicer tests..

bobleesj:
@sbillinge Also passes locally now. Just checked out your branch.

sbillinge:
thank for the review @bobleesj .  I will merge this now so the team can push on.....

Issue #194, opened 2024-12-04
overload the __setter__ in DiffractionObject to correctly update arrays


sbillinge:
### Problem

the `all_arrays` attribute can be updated with anything right now.    this could inadvertently lead to invalid arrays.

### Proposed solution
We would like to block this so it can't be updated (issue a warning?) and these arrays only updated by `insert_scattering_object`

do this by defining an @property for this attribute.

There may also be some other behaviors we want to modify to ensure valid `__set__` behavior

PR #196, opened 2024-12-04
Use `@property` to prevent direct modification of `all_arrays`

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/194

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -181,6 +181,17 @@ class DiffractionObject:
             divided.on_q[1] = other.on_q[1] / self.on_q[1]
         return divided
 
+    @property
+    def all_arrays(self):
+        return self._all_arrays
+
+    @all_arrays.setter
+    def all_arrays(self, value):
+        raise AttributeError(
+            "Direct modification of 'all_arrays' is not allowed."
+            "Please use 'insert_scattering_quantity' to modify it."
+        )
+
     def set_angles_from_list(self, angles_list):
         self.angles = angles_list
         self.n_steps = len(angles_list) - 1.0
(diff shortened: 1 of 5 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -181,6 +181,17 @@ def __rtruediv__(self, other):
             divided.on_q[1] = other.on_q[1] / self.on_q[1]
         return divided
 
+    @property
+    def all_arrays(self):
+        return self._all_arrays
```
bobleesj, line 186: A private property `_all_arrays` is used internally.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -181,6 +181,17 @@ def __rtruediv__(self, other):
...
+
+    @all_arrays.setter
```
bobleesj, line 188: when the user attempts to modify `all_array`, throw an exception. I've thought it might be better to use `exception` rather than `warning` since our `_set_xarrays` is doing some heavy lifting work and we want our users to use the way we want them to do?

tests/test_diffraction_objects.py
```diff
@@ -341,5 +341,27 @@ def test_dump(tmp_path, mocker):
 
 @pytest.mark.parametrize("inputs, expected", tc_params)
 def test_constructor(inputs, expected):
-    actualdo = DiffractionObject(**inputs)
-    compare_dicts(actualdo.__dict__, expected)
+    actual_do = DiffractionObject(**inputs)
```
bobleesj, line 344: 1) This is to ensure we get `all_arrays` instead of `_all_arrays` and 2) we want to mimic the way the user extracts properties - one by one.

tests/test_diffraction_objects.py
```diff
@@ -341,5 +341,27 @@ def test_dump(tmp_path, mocker):
...
+        "tthmin": actual_do.tthmin,
+        "tthmax": actual_do.tthmax,
+        "dmin": actual_do.dmin,
+        "dmax": actual_do.dmax,
+        "wavelength": actual_do.wavelength,
+    }
+    compare_dicts(actual_dict, expected)
+
+
+def test_all_array_setter():
```
bobleesj, line 362: Test for throwing an exception when the user attempts to modify `all_arrays` directly.

bobleesj (commented):
@sbillinge Ready for review!

tests/test_diffraction_objects.py
```diff
@@ -341,5 +341,27 @@ def test_dump(tmp_path, mocker):
 
 @pytest.mark.parametrize("inputs, expected", tc_params)
 def test_constructor(inputs, expected):
-    actualdo = DiffractionObject(**inputs)
-    compare_dicts(actualdo.__dict__, expected)
+    actual_do = DiffractionObject(**inputs)
+    actual_dict = {
```
sbillinge, line 345: we normally do this work using `actual_do.__dict__` so the test will fail if someone adds a new attribute to `DiffractionObject`

tests/test_diffraction_objects.py
```diff
@@ -341,5 +341,27 @@ def test_dump(tmp_path, mocker):
...
+        "wavelength": actual_do.wavelength,
+    }
+    compare_dicts(actual_dict, expected)
+
+
+def test_all_array_setter():
+    actual_do = DiffractionObject()
+
+    # Attempt to directly modify the property
+    with pytest.raises(AttributeError, match="Direct modification of 'all_arrays' is not allowed."):
```
sbillinge, line 366: let's follow our favorite pattern.  "what went wrong.  How to fix it".  The first part is done (though I would add the word "attribute" before 'all_arrays') but not the second?  I think we want to point them towards using `insert_diffraction_object` or sthg, but give it some thought.

sbillinge (commented):
pls see inline.  I mostly just reviewed the tests.

tests/test_diffraction_objects.py
```diff
@@ -341,5 +342,36 @@ def test_dump(tmp_path, mocker):
 
 @pytest.mark.parametrize("inputs, expected", tc_params)
 def test_constructor(inputs, expected):
-    actualdo = DiffractionObject(**inputs)
-    compare_dicts(actualdo.__dict__, expected)
```
bobleesj, line 345: Can we use a library `DeepDiff` to compare two dicts instead? We've manually defined `compare_dicts` and `dicts_equal` within `test_diffraction_objects.py` which can be replaced using `DeepDiff` across all .py files.

tests/test_diffraction_objects.py
```diff
@@ -341,5 +342,36 @@ def test_dump(tmp_path, mocker):
...
+    actual_do = DiffractionObject(**inputs)
+    diff = DeepDiff(actual_do.__dict__, expected, ignore_order=True, significant_digits=4)
+    # Ensure there is no difference
+    assert diff == {}
+
+
+def test_all_array_getter():
```
bobleesj, line 351: Test getter for `all_arrays`

bobleesj:
@sbillinge ready for review again

tests/test_diffraction_objects.py
bobleesj: yes, done.

tests/test_diffraction_objects.py
bobleesj: yes done.

PR #197, opened 2024-12-05
d_to_q, q_to_d

### LINKED ISSUE #176: diffraction objects implement d_to_q, q_to_d
### Problem

We want to have functions to convert between qspace and dspace.

### Proposed solution
Refer to #154 for previous discussions. Need to handle cases when q or d=0.

yucongalicechen:
closes #176 
@sbillinge ready for review

### DIFF
--- a/src/diffpy/utils/transforms.py
+++ b/src/diffpy/utils/transforms.py
@@ -16,6 +16,9 @@ invalid_q_or_wavelength_emsg = (
     "The supplied q-array and wavelength will result in an impossible two-theta. "
     "Please check these values and re-instantiate the DiffractionObject with correct values."
 )
+invalid_input_emsg = (
+    "Input values have resulted in an infinite output. Please ensure there are no zeros in the input."
+)
 
 
 def _validate_inputs(q, wavelength):
@@ -58,7 +61,7 @@ def q_to_tth(q, wavelength):
     -------
     tth : 1D array
         The array of :math:`2\theta` values in degrees numpy.array([tths]).
-        This is the correct format for loading into diffpy.utils.DiffractionOject.on_tth
+        This is the correct format for loading into diffpy.utils.DiffractionObject.on_tth.
     """
     _validate_inputs(q, wavelength)
     q.astype(float)
@@ -92,9 +95,8 @@ def tth_to_q(tth, wavelength):
 
     Parameters
     ----------
-    tth : 2D array
-        The array of :math:`2\theta` values and :math: 'i' intensity values, np.array([[tths], [is]]).
-        This is the same format as, and so can accept, diffpy.utils.DiffractionOject.on_tth
+    tth : 1D array
+        The array of :math:`2\theta` values np.array([tths]).
         The units of tth are expected in degrees.
 
     wavelength : float
@@ -102,11 +104,10 @@ def tth_to_q(tth, wavelength):
 
     Returns
     -------
-    on_q : 2D array
-        The array of :math:`q` values and :math: 'i' intensity values unchanged,
-        np.array([[qs], [is]]).
+    q : 1D array
+        The array of :math:`q` values np.array([qs]).
         The units for the q-values are the inverse of the units of the provided wavelength.
-        This is the correct format for loading into diffpy.utils.DiffractionOject.on_q
+        This is the correct format for loading into diffpy.utils.DiffractionObject.on_q.
     """
     tth.astype(float)
     if np.any(np.deg2rad(tth) > np.pi):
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/transforms.py
```diff
@@ -58,7 +61,7 @@ def q_to_tth(q, wavelength):
     -------
     tth : 1D array
         The array of :math:`2\theta` values in degrees numpy.array([tths]).
-        This is the correct format for loading into diffpy.utils.DiffractionOject.on_tth
+        This is the correct format for loading into diffpy.utils.DiffractionObject.on_tth.
```
yucongalicechen, line 64: should we delete this from docstring, since we're oprating on 1d array now?

tests/test_transforms.py
```diff
@@ -96,3 +96,73 @@ def test_tth_to_q(inputs, expected):
...
         tth_to_q(inputs[1], inputs[0])
+
+
+params_q_to_d = [
+    # UC1: User specified empty q values
+    ([np.array([])], np.array([])),
+    # UC2: User specified valid q values
+    (
+        [np.array([np.pi / 6, 1 * np.pi, 2 * np.pi, 3 * np.pi, 4 * np.pi, 5 * np.pi])],
+        np.array([0.4, 0.5, 0.66667, 1, 2, 12]),
```
yucongalicechen, line 107: shall we print a warning message if d is too small / large? 12 might not be very extreme here, but if we have q~0 we might get a really large d. i think it doesn't really make sense if we change that d value, but just warn the users that we get a value that may cause trouble.

src/diffpy/utils/transforms.py
```diff
@@ -16,6 +16,9 @@
     "The supplied q-array and wavelength will result in an impossible two-theta. "
     "Please check these values and re-instantiate the DiffractionObject with correct values."
 )
+invalid_input_emsg = (
```
sbillinge, line 19: I think zero's are valid.  I could imagine many situations where I would like to give a Q all the way to zero, so I am not sure I like this.

src/diffpy/utils/transforms.py
```diff
@@ -92,21 +95,19 @@ def tth_to_q(tth, wavelength):
...
     Returns
     -------
-    on_q : 2D array
-        The array of :math:`q` values and :math: 'i' intensity values unchanged,
-        np.array([[qs], [is]]).
+    q : 1D array
+        The array of :math:`q` values np.array([qs]).
         The units for the q-values are the inverse of the units of the provided wavelength.
-        This is the correct format for loading into diffpy.utils.DiffractionOject.on_q
+        This is the correct format for loading into diffpy.utils.DiffractionObject.on_q.
```
sbillinge, line 110: delete, as discussed.

src/diffpy/utils/transforms.py
```diff
@@ -121,17 +122,49 @@ def tth_to_q(tth, wavelength):
...
+        The array of :math:`q` values np.array([qs]).
+        The units of q must be reciprocal of the units of wavelength.
+
+    Returns
+    -------
+    d : 1D array
+        The array of :math:`d` values np.array([ds]).
+    """
+    if 0 in q:
+        raise ValueError(invalid_input_emsg)
```
sbillinge, line 141: just make this `np.inf` but also give a warning message that the largest d is infinite and can't be plotted?  Something like that?

src/diffpy/utils/transforms.py
```diff
@@ -121,17 +122,49 @@ def tth_to_q(tth, wavelength):
...
+    return 2.0 * np.pi / copy(q)[::-1]
```
sbillinge, line 142: what does this (`[::-1]`) do?  why not just `2.0 * np.pi / copy(q)`?

src/diffpy/utils/transforms.py
```diff
@@ -121,17 +122,49 @@ def tth_to_q(tth, wavelength):
...
+    d : 1D array
+        The array of :math:`d` values np.array([ds]).
+
+    Returns
+    -------
+    q : 1D array
+        The array of :math:`q` values np.array([qs]).
+        The units of q must be reciprocal of the units of wavelength.
+    """
+    if 0 in d:
```
sbillinge, line 165: see discussion above

tests/test_transforms.py
sbillinge: no I don't think so.

Has the d-array become inverted?  Shouldn't it go from large to small?

sbillinge (commented):
please see comments

yucongalicechen:
@sbillinge updated. please review again.

src/diffpy/utils/transforms.py
```diff
@@ -16,9 +16,7 @@
     "The supplied q-array and wavelength will result in an impossible two-theta. "
     "Please check these values and re-instantiate the DiffractionObject with correct values."
 )
-invalid_input_emsg = (
-    "Input values have resulted in an infinite output. Please ensure there are no zeros in the input."
-)
+inf_output_msg = "WARNING: The largest output is infinite and cannot be plotted."
```
bobleesj, line 19: @sbillinge out of curiosity, 

for both error and warning messages, do we still want to maintain our convention of 1) reason for error, (2) what to do to fix it ?

sbillinge: I tweaked this a bit and downgraded it to an INFO from a WARNING....don't want users to freak out....

PR #199, opened 2024-12-06
add d_to_tth, tth_to_d

### LINKED ISSUE #175: diffraction objects implement d_to_tth, tth_to_d
### Problem

We don't have these functions.

### Proposed solution
Refer to #154 for previous progress. Still need to handle the case when d=0.

yucongalicechen:
closes #175
@sbillinge ready for review

### DIFF
--- a/src/diffpy/utils/transforms.py
+++ b/src/diffpy/utils/transforms.py
@@ -12,8 +12,8 @@ wavelength_warning_emsg = (
     "and number is the wavelength in angstroms."
 )
 invalid_tth_emsg = "Two theta exceeds 180 degrees. Please check the input values for errors."
-invalid_q_or_wavelength_emsg = (
-    "The supplied q-array and wavelength will result in an impossible two-theta. "
+invalid_q_or_d_or_wavelength_emsg = (
+    "The supplied input array and wavelength will result in an impossible two-theta. "
     "Please check these values and re-instantiate the DiffractionObject with correct values."
 )
 inf_output_wmsg = "WARNING: The largest output is infinite and cannot be plotted."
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/transforms.py
```diff
@@ -12,10 +12,11 @@
     "and number is the wavelength in angstroms."
 )
 invalid_tth_emsg = "Two theta exceeds 180 degrees. Please check the input values for errors."
-invalid_q_or_wavelength_emsg = (
-    "The supplied q-array and wavelength will result in an impossible two-theta. "
+invalid_q_or_d_or_wavelength_emsg = (
+    "The supplied input array and wavelength will result in an impossible two-theta. "
```
yucongalicechen, line 16: changing this to input so that we can use that for d as well

sbillinge:
@yucongalicechen this looks good, but I made a small tweak to an error message in the other PR so this is now conflicted.  Please can you merge main, keeping the new message, to resolve this, and then I can merge it.

tests/test_transforms.py
```diff
@@ -103,8 +103,8 @@ def test_tth_to_q_bad(inputs, expected):
     ([np.array([])], np.array([])),
     # UC2: User specified valid q values
     (
-        [np.array([5 * np.pi, 4 * np.pi, 3 * np.pi, 2 * np.pi, np.pi, 0])],
-        np.array([0.4, 0.5, 0.66667, 1, 2, np.inf]),
+        [np.array([0, 1 * np.pi, 2 * np.pi, 3 * np.pi, 4 * np.pi, 5 * np.pi])],
+        np.array([np.inf, 2, 1, 0.66667, 0.5, 0.4]),
```
yucongalicechen, line 107: small edit so that we have an increasing qarray

PR #204, opened 2024-12-07
Add `copy()` method for `DiffractionObject`

### LINKED ISSUE #201: add a copy method to diffraction objects
### Problem
That would be nice.  Something like 
```
def copy(self):
    return DiffractionObject(**self.everything)
```
where teat this as pseudocode....it is supposed to show intent, not how it would actually be coded.

usage would be that a test like this would pass:
```
copy_of_DO = DO.copy()
assert copy_of_DO == DO
```

bobleesj:
Closes #201 - add `copy()` for `DiffracitonObject`

### DIFF
--- a/tests/test_diffraction_objects.py
+++ b/tests/test_diffraction_objects.py
@@ -410,3 +410,15 @@ def test_all_array_setter():
         "Please use 'insert_scattering_quantity' to modify `all_arrays`.",
     ):
         actual_do.all_arrays = np.empty((4, 4))
+
+
+def test_copy_object():
+    do = DiffractionObject(
+        name="test",
+        wavelength=4.0 * np.pi,
+        xarray=np.array([0.0, 90.0, 180.0]),
+        yarray=np.array([1.0, 2.0, 3.0]),
+        xtype="tth",
+    )
+    copy_of_DO = do.copy()
+    assert do == copy_of_DO
--- a/tests/test_diffraction_objects.py
+++ b/tests/test_diffraction_objects.py
@@ -420,7 +420,6 @@ def test_copy_object():
         yarray=np.array([1.0, 2.0, 3.0]),
         xtype="tth",
     )
-    copy_of_DO = do.copy()
-    assert do == copy_of_DO
-    assert id(do) != id(copy_of_DO)
-
+    do_copy = do.copy()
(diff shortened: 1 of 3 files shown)


tests/test_diffraction_objects.py
```diff
@@ -410,3 +410,15 @@ def test_all_array_setter():
...
+def test_copy_object():
+    do = DiffractionObject(
+        name="test",
+        wavelength=4.0 * np.pi,
+        xarray=np.array([0.0, 90.0, 180.0]),
+        yarray=np.array([1.0, 2.0, 3.0]),
+        xtype="tth",
+    )
+    copy_of_DO = do.copy()
+    assert do == copy_of_DO
```
bobleesj, line 424: Utizling the object's `__eq__` nice implementation

bobleesj:
@sbillinge ready for review

tests/test_diffraction_objects.py
sbillinge: To complete this test also assert id(actual) != Id(initial)

sbillinge: Then the copy is not pointing to the same object

sbillinge (commented):
Pls see inline

bobleesj:
@sbillinge Right! Thank you. Additional test added.

tests/test_diffraction_objects.py
bobleesj: yes done below.

bobleesj:
@sbillinge ready for review - just enhanced the variable name from `copy_of_DO` to `do_copy` for consistency.

sbillinge:
@bobleesj not sure why codecov is complaining here...diffpy has a global org secret token.  Maybe it is being flaky.  I will just try and rerun it?  after than an empty commit?

bobleesj:
Screenshot 2024-12-08 at 10 05 26 AM

Codecov looks okay to me in fact 0.05% has gone up! @sbillinge

sbillinge:
> Screenshot 2024-12-08 at 10 05 26 AM
> 
> Codecov looks okay to me in fact 0.05% has gone up! @sbillinge

it failed to upload so failed.  I reran it and it uploaded and passed.....   CC is officially flaky....

PR #206, opened 2024-12-07
Match `tests` folder structure to `src` folder structure

### LINKED ISSUE #205: Match `tests` folder structure to `src` folder structure
### Problem

src:

Screenshot 2024-12-07 at 4 54 40 PM

tests:

Screenshot 2024-12-07 at 4 54 56 PM

### Proposed solution

Create `parsers` and `wx` folders within `tests` and move relevant files.

bobleesj:
Closes #205 - match `tests`  and `src` folder structures.


bobleesj:
@sbillinge ready for review, just quick cosmetics fix.

sbillinge:
What is the reason for doing this?

bobleesj:
> What is the reason for doing this?

Since test files are all located under `tests`, I am trying  to mirror the structure of the `src` directory with the `tests` directory. 

```
src/
└── parser/
    └── loaddata.py

tests/
└── parser/
    └── test_loaddata.py
```

sbillinge:
> > What is the reason for doing this?
> 
> Since test files are all located under `tests`, I am trying to mirror the structure of the `src` directory with the `tests` directory.
> 
> ```
> src/
> └── parser/
>     └── loaddata.py
> 
> tests/
> └── parser/
>     └── test_loaddata.py
> ```

Yes, I understand what you are trying to do.  I am wondering why.  I would think just having the tests flat is easier to manage and maintain.  They don't have to replicate any package structure.

sbillinge:
if we worry about name clashes (there shouldn't be too many) it should be easy to disambiguate.  If we thought it helped we could use the full package name in the file name if we liked too, like `test_parser.loaddata.py` but even this coud be a bit of overkill.  But I could be persuaded otherwise.

bobleesj:
> understand what you are trying to do. I am wondering why. I would think just having the tests flat is easier to manage and maintain. They don't have to replicate any package structure.

I had thought it was standard to mirror the src structure, but now, after thinking and searching more, I don't have a clear reason. Also, adopting a flat structure doesn’t deviate from the fact that tests aren't written for the code but rather for flexibility, and we write tests for intended behavior, not for the code itself.

PR #207, opened 2024-12-07
add deprecation warning to resample function

### LINKED ISSUE #139: deprecate legacy resampler?
### Problem

there is a legacy resampler in parsers.  @Sparks29032 was this from before or did you write this then write a better version.   In general, keeping around things that are not used is just technical debt and we should get rid of it. 

### Proposed solution

 If it is something from Pavol days we may want to keep it as other code is likely using it.  If we think the code that is using it should start using the wsinterp instead we should deprectate it which means that when it is called we print a warning message that it is deprecated and will be removed in a version two mid-bumps later but should be replaced with wsinterp, then make issues linked to two bumps later to remove it.

alisnwu:
closes #139

as per discussion in #180 warnings.warn passes pytest under python 3.13, 3.12, 3.11

@sbillinge @bobleesj ready for review

Screenshot 2024-12-07 at 5 16 14 PM

### DIFF
--- a/news/resampler-dep-warn.rst
+++ b/news/resampler-dep-warn.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Deprecation warning for resample function in resampler
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/resampler-dep-warn.rst
+++ b/news/resampler-dep-warn.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* Deprecation warning for resample function in resampler
+* <news item>
 
 **Changed:**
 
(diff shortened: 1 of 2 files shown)


news/resampler-dep-warn.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* Deprecation warning for resample function in resampler
```
sbillinge, line 3: we are not adding functionality so move this from here to `Deprecated`

sbillinge (commented):
please see inline

alisnwu:
@sbillinge fixed news item, ready for review

news/resampler-dep-warn.rst
```diff
@@ -1,14 +1,14 @@
...
+* <news item>
 
 **Changed:**
 
 * <news item>
 
 **Deprecated:**
 
-* <news item>
+* Deprecation warning for resample function in resampler
```
sbillinge, line 11: how about "`resample` function in `resampler`.  Repaced with `wsinterp`" or sthg like that?

This news is user-facing so it should be targeted at someone who might be using the code.

sbillinge (commented):
pls see inline.

sbillinge:
@alisnwu why are you closing this?

PR #208, opened 2024-12-07
More info on updating global config for users

### LINKED ISSUE #156: New workflow for users to update the global config file
### Problem

Need a better workflow for users to know how to update the global config file easily.

### Proposed solution
Current instructions are in documentation for diffpy.utils so maybe we can include some message that gives instructions or points to the documentation

alisnwu:
closes #156 

Screenshot 2024-12-07 at 5 52 42 PM

@sbillinge ready for review, new message points to where users can update manually and examples on the diffpy.utils link

### DIFF
--- a/news/config-UX.rst
+++ b/news/config-UX.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* additional information to users to relieve frustration in finding how to update global config
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/config-UX.rst
+++ b/news/config-UX.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* additional information to users to relieve frustration in finding how to update global config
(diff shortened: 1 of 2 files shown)


news/config-UX.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* additional information to users to relieve frustration in finding how to update global config
```
sbillinge, line 3: no functionality added.  Move this to `Fixed`

sbillinge (commented):
please see inline.  news needs fixing but otherwise this is great

alisnwu:
@sbillinge fixed news item, ready for review

sbillinge:
btw @alisnwu I closed this before reading https://www.diffpy.org/diffpy.utils/examples/toolsexample.html but there I saw no clear information about how to update the default user information.  I would put it as an enumerated list to make it clearer.  there are at least 4 different ways of setting these things with more or less permanence.  Make it as an enumerated list.  But also discuss with and example how to updage your default file using a text editor.

Also, is this a typo?
```
from diffpy.utils.tools import get_user_info
package_metadata = get_package_info("my_package")
```

sbillinge:
moved this comment to issue #210

PR #209, opened 2024-12-08
Check valid x and y array lenghts and xtypes in `insert_scattering_quantity` 

### LINKED ISSUE #200: insert_scattering_quantity check valid arrays and xtypes
### Proposed solution

In the `insert_scattering_quantity` function, we want to add these features: (1) check if xarray is empty or not, if not check that yarray has the same length as xarray and xtype is valid (one of XQUANTITIES), (2) set attribute so that we don't allow setting an invalid xtype.

bobleesj:
Closes #200

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -198,12 +198,12 @@ class DiffractionObject:
         raise AttributeError(_setter_wmsg("all_arrays"))
 
     @property
-    def xtype(self):
-        return self._xtype
+    def input_xtype(self):
+        return self._input_xtype
 
-    @xtype.setter
-    def xtype(self, _):
-        raise AttributeError(_setter_wmsg("xtype"))
+    @input_xtype.setter
+    def input_xtype(self, _):
+        raise AttributeError(_setter_wmsg("input_xtype"))
 
     def set_angles_from_list(self, angles_list):
         self.angles = angles_list
@@ -289,7 +289,7 @@ class DiffractionObject:
         """
 
         if xtype is None:
-            xtype = self._xtype
+            xtype = self._input_xtype
         array = self.on_xtype(xtype)[0]
         if len(array) == 0:
             raise ValueError(f"The '{xtype}' array is empty. Please ensure it is initialized.")
@@ -348,7 +348,7 @@ class DiffractionObject:
         """
         self._set_xarrays(xarray, xtype)
         self._all_arrays[:, 0] = yarray
-        self._xtype = xtype
+        self._input_xtype = xtype
         # only update these optional values if non-empty quantities are passed to avoid overwriting
         # valid data inadvertently
         if metadata:
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -26,6 +26,13 @@ def _xtype_wmsg(xtype):
     )
 
 
+def _setter_wmsg(attribute):
+    return (
+        f"Direct modification of attribute '{attribute}' is not allowed."
+        f"Please use 'insert_scattering_quantity' to modify '{attribute}'.",
+    )
+
+
 class DiffractionObject:
     def __init__(
         self, name=None, wavelength=None, scat_quantity=None, metadata=None, xarray=None, yarray=None, xtype=None
@@ -45,6 +52,7 @@ class DiffractionObject:
             xarray = np.empty(0)
         if yarray is None:
             yarray = np.empty(0)
+
         self.insert_scattering_quantity(xarray, yarray, xtype)
 
     def __eq__(self, other):
@@ -186,11 +194,16 @@ class DiffractionObject:
         return self._all_arrays
 
     @all_arrays.setter
-    def all_arrays(self, value):
-        raise AttributeError(
-            "Direct modification of attribute 'all_arrays' is not allowed."
-            "Please use 'insert_scattering_quantity' to modify `all_arrays`."
-        )
+    def all_arrays(self, _):
+        raise AttributeError(_setter_wmsg("all_arrays"))
+
+    @property
+    def xtype(self):
+        return self._xtype
+
+    @xtype.setter
+    def xtype(self, _):
+        raise AttributeError(_setter_wmsg("xtype"))
 
     def set_angles_from_list(self, angles_list):
         self.angles = angles_list
@@ -261,7 +274,7 @@ class DiffractionObject:
 
     def get_array_index(self, value, xtype=None):
         """
-        returns the index of the closest value in the array associated with the specified xtype
+        Return the index of the closest value in the array associated with the specified xtype.
 
         Parameters
         ----------
@@ -276,7 +289,7 @@ class DiffractionObject:
         """
 
         if xtype is None:
-            xtype = self.input_xtype
+            xtype = self._xtype
         array = self.on_xtype(xtype)[0]
         if len(array) == 0:
             raise ValueError(f"The '{xtype}' array is empty. Please ensure it is initialized.")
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -186,11 +194,16 @@ def all_arrays(self):
...
+    def all_arrays(self, _):
+        raise AttributeError(_setter_wmsg("all_arrays"))
+
+    @property
+    def xtype(self):
+        return self._xtype
+
+    @xtype.setter
+    def xtype(self, _):
+        raise AttributeError(_setter_wmsg("xtype"))
```
bobleesj, line 206: @sbillinge Like Would you also like to prevent the user from directly modifying `xtype`?

src/diffpy/utils/diffraction_objects.py
```diff
@@ -276,7 +289,7 @@ def get_array_index(self, value, xtype=None):
         """
 
         if xtype is None:
-            xtype = self.input_xtype
+            xtype = self._xtype
```
bobleesj, line 292: If we do not want the user from directly modifying `xtype`, like `_all_arrays`, `_xtype` is used internally.

tests/test_diffraction_objects.py
```diff
@@ -407,6 +406,23 @@ def test_all_array_setter():
...
         match="Direct modification of attribute 'all_arrays' is not allowed."
-        "Please use 'insert_scattering_quantity' to modify `all_arrays`.",
+        "Please use 'insert_scattering_quantity' to modify 'all_arrays'.",
     ):
         actual_do.all_arrays = np.empty((4, 4))
+
+
+def test_xtype_getter():
+    do = DiffractionObject(xtype="tth")
+    assert do.xtype == "tth"
```
bobleesj, line 416: Test getter for `xtype`

tests/test_diffraction_objects.py
```diff
@@ -407,6 +406,23 @@ def test_all_array_setter():
...
+
+
+def test_xtype_setter():
```
bobleesj, line 419: Test setter for `xtype` - not allowed

bobleesj:
@yucongalicechen could you please give me a quick review? Since you've written the issue, I want to ensure this is what was imagined.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -347,12 +360,25 @@ def insert_scattering_quantity(
...
+            raise ValueError(
+                "`xarray` and `yarray` must have the same length. "
+                "Please re-initialize `DiffractionObject` or re-run the method `insert_scattering_quantity` "
+                "with `xarray` and `yarray` of identical length."
+            )
+
+        # Check xtype is valid. An empty string is the default value.
+        if xtype != "":
+            if xtype not in XQUANTITIES:
+                raise ValueError(_xtype_wmsg(xtype))
```
bobleesj, line 374: If `xtype` is not one of the `XQUANTITIES` values, throw a `ValueError`

src/diffpy/utils/diffraction_objects.py
yucongalicechen: I like how you changed input_xtype to xtype! I do think people can directly modify xtype though (@sbillinge please confirm). For example I might want to plot my diffraction object on tth first, and then change its xtype to q so I can visualize it in qspace.

sbillinge: @yucongalicechen that is not how it works.  the attribute was called `input_xtype` because we wanted to keep a record of the xtype input by the user.  So I don't want to change its name.  When xtype is used for plotting any xtype can be given, for example `diffobj.on_xtype('q')` would satisfy @yucongalicechen 's UC even if the data were input on a `tth` grid.

Since we only use it internally, we could change the attribute name to `_input_xtype` to make it private.  I think it should only be set by `insert_data` and not re-settable.

bobleesj: @sbillinge  - done. `input_type` is used internally and is it only gettable.

bobleesj:
> Since we only use it internally, we could change the attribute name to `_input_xtype` to make it private. I think it should only be set by `insert_data` and not re-settable.


@sbillinge Ready for review. Once this is merged, I can change the method name from `insert_scattering_quantity` to `input_data`?

tests/test_diffraction_objects.py
```diff
@@ -406,12 +405,39 @@ def test_all_array_setter():
...
+    ):
+        DiffractionObject(xarray=np.array([1.0, 2.0]), yarray=np.array([0.0, 0.0, 0.0]))
+
+
+def test_input_xtype_getter():
+    do = DiffractionObject(xtype="tth")
+    assert do.input_xtype == "tth"
+
+
+def test_input_xtype_setter():
```
bobleesj, line 429: Not allowed to reset `_input_data`

bobleesj:
> Since we only use it internally, we could change the attribute name to _input_xtype to make it private. I think it should only be set by insert_data and not re-settable.

ready for review - @sbillinge

sbillinge (commented):
Sorry, this disappeared of my notifications but I never submitted the review

src/diffpy/utils/diffraction_objects.py
```diff
@@ -347,12 +369,17 @@ def insert_scattering_quantity(
         if wavelength is not None:
             self.wavelength = wavelength
 
+        # Check xtype is valid. An empty string is the default value.
+        if xtype != "":
```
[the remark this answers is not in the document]
bobleesj: From our current init implementation, if `xtype` is not provided for `DiffractionObjects`, then `xtype` becomes an empty string as shown here:

```python
class DiffractionObject:
    def __init__(
        self, name=None, wavelength=None, scat_quantity=None, metadata=None, xarray=None, yarray=None, xtype=None
    ):
        ...
        self.metadata = metadata
        if xtype is None:
            xtype = ""
```

Hence, we need to have `if xtype != ""` to ensure that we do not provide the warning msg when no value is provided for `xtype`.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -186,11 +194,16 @@ def all_arrays(self):
...
     @all_arrays.setter
-    def all_arrays(self, value):
-        raise AttributeError(
-            "Direct modification of attribute 'all_arrays' is not allowed."
-            "Please use 'insert_scattering_quantity' to modify `all_arrays`."
-        )
+    def all_arrays(self, _):
+        raise AttributeError(_setter_wmsg("all_arrays"))
+
+    @property
```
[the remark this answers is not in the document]
bobleesj: Yes,

```python
    @input_xtype.setter
    def input_xtype(self, _):
        raise AttributeError(_setter_wmsg("input_xtype"))
```

This code basically, gives `AttributeError` when the user attempts to modify `input_xtype` manually.

bobleesj (commented):
@sbillinge ready for review - didn't modify any code but replied to your in-line comments.

src/diffpy/utils/diffraction_objects.py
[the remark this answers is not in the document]
sbillinge: I see.  This is what I am seeing on my screen which is why I thought otherwise, but it may not be showing the latest code on GH?
image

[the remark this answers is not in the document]
sbillinge: not looking far enough down.  Thanks for the explanation

src/diffpy/utils/diffraction_objects.py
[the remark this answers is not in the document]
sbillinge: That makes sense.  I guess this is left over from our/my desire to allow people to instantiate an empty DO and then fill it later.  Let's leave it like that, but we may be accumulating annoying technical debt that we wouldn't have if we went away from that choice.

sbillinge (commented):
Please see comments.  It all looks good, modulo a nasty feeling that we may have made an imperfect choice....

PR #211, opened 2024-12-08
tests for scale_to

### LINKED ISSUE #49: diffraction object
- [x] carry around resampled data and list resamples
- [x] plot data on chosen resample
- [ ]  automatically scale data, for example to a point or to a maximum. e.g., `mo.scale_to(do.on_tth, xvalue=52)`
- [ ]  extend (3) with an offset e.g., `mo.scale_to(do.on_tth, xvalue=52).plus(5)`

### LINKED ISSUE #186: write tests for

yucongalicechen:
closes #49, closes #186

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -388,11 +388,9 @@ class DiffractionObject:
         -------
         the rescaled DiffractionObject as a new object
         """
-        scaled = deepcopy(self)
+        scaled = self.copy()
         xtype = "q" if q is not None else "tth" if tth is not None else "d" if d is not None else "q"
         data, target = self.on_xtype(xtype), target_diff_object.on_xtype(xtype)
-        if len(data[0]) == 0 or len(target[0]) == 0:
-            raise ValueError("I cannot scale diffraction objects with empty arrays.")
 
         xvalue = q if xtype == "q" else tth if xtype == "tth" else d
         if xvalue is None:
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -397,15 +397,14 @@ class DiffractionObject:
 
         The y-value in the target at the closest specified x-value will be used as the factor to scale to.
         The entire array is scaled by this factor so that one object places on top of the other at that point.
-        If multiple values of `q`, `tth`, or `d` are provided, an error will be raised.
-        If none are provided, the midpoint of the current object's `q`-array will be used.
+        If multiple values of `q`, `tth`, or `d` are provided, or none are provided, an error will be raised.
 
         Parameters
         ----------
         target_diff_object: DiffractionObject
             the diffraction object you want to scale the current one onto
 
-        q, tth, d : float, optional, default is the midpoint of the current object's `q`-array
+        q, tth, d : float, optional, must specify exactly one of them
             the xvalue (in `q`, `tth`, or `d` space) to align the current and target objects
 
         offset : float, optional, default is 0
@@ -417,19 +416,19 @@ class DiffractionObject:
         """
         scaled = self.copy()
         count = sum([q is not None, tth is not None, d is not None])
-        if count > 1:
-            raise ValueError("You can only specify one of 'q', 'tth', or 'd'. Please rerun specifying only one.")
+        if count != 1:
+            raise ValueError(
+                "You must specify exactly one of 'q', 'tth', or 'd'. Please rerun specifying only one."
+            )
 
         xtype = "q" if q is not None else "tth" if tth is not None else "d" if d is not None else "q"
         data, target = self.on_xtype(xtype), target_diff_object.on_xtype(xtype)
 
         xvalue = q if xtype == "q" else tth if xtype == "tth" else d
-        if xvalue is None:
-            xvalue = (data[0][0] + data[0][-1]) / 2.0
 
-        x_data, x_target = (np.abs(data[0] - xvalue)).argmin(), (np.abs(target[0] - xvalue)).argmin()
-        y_data, y_target = data[1][x_data], target[1][x_target]
-        scaled._all_arrays[:, 0] = data[1] * y_target / y_data + offset
+        xindex_data = (np.abs(data[0] - xvalue)).argmin()
+        xindex_target = (np.abs(target[0] - xvalue)).argmin()
+        scaled._all_arrays[:, 0] = data[1] * target[1][xindex_target] / data[1][xindex_data] + offset
         return scaled
 
     def on_xtype(self, xtype):
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -397,7 +397,7 @@ class DiffractionObject:
 
         The y-value in the target at the closest specified x-value will be used as the factor to scale to.
         The entire array is scaled by this factor so that one object places on top of the other at that point.
-        If multiple values of `q`, `tth`, or `d` are provided, the priority is `q` > `tth` > `d`.
+        If multiple values of `q`, `tth`, or `d` are provided, an error will be raised.
         If none are provided, the midpoint of the current object's `q`-array will be used.
 
         Parameters
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -366,7 +366,7 @@ class DiffractionObject:
 
     def scale_to(self, target_diff_object, xtype=None, xvalue=None):
         f"""
-        returns a new diffraction object which is the current object but recaled in y to the target
+        returns a new diffraction object which is the current object but rescaled in y to the target
 
         Parameters
         ----------
@@ -390,14 +390,15 @@ class DiffractionObject:
 
         data = self.on_xtype(xtype)
         target = target_diff_object.on_xtype(xtype)
+        if len(data[0]) == 0 or len(target[0]) == 0 or len(data[0]) != len(target[0]):
+            raise ValueError("I cannot scale two diffraction objects with empty or different lengths.")
         if xvalue is None:
             xvalue = data[0][0] + (data[0][-1] - data[0][0]) / 2.0
 
         xindex = (np.abs(data[0] - xvalue)).argmin()
         ytarget = target[1][xindex]
         yself = data[1][xindex]
-        scaled.on_tth[1] = data[1] * ytarget / yself
-        scaled.on_q[1] = data[1] * ytarget / yself
+        scaled._all_arrays[:, 0] = data[1] * ytarget / yself
         return scaled
 
     def on_xtype(self, xtype):
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -405,7 +405,8 @@ class DiffractionObject:
             the diffraction object you want to scale the current one onto
 
         q, tth, d : float, optional, must specify exactly one of them
-            the xvalue (in `q`, `tth`, or `d` space) to align the current and target objects
+            The value of the x-array where you want the curves to line up vertically.
+            Specify a value on one of the allowed grids, q, tth, or d), e.g., q=10.
 
         offset : float, optional, default is 0
             an offset to add to the scaled y-values
@@ -421,8 +422,9 @@ class DiffractionObject:
                 "You must specify exactly one of 'q', 'tth', or 'd'. Please rerun specifying only one."
             )
 
-        xtype = "q" if q is not None else "tth" if tth is not None else "d" if d is not None else "q"
-        data, target = self.on_xtype(xtype), target_diff_object.on_xtype(xtype)
+        xtype = "q" if q is not None else "tth" if tth is not None else "d"
+        data = self.on_xtype(xtype)
+        target = target_diff_object.on_xtype(xtype)
 
         xvalue = q if xtype == "q" else tth if xtype == "tth" else d
 
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -364,41 +364,43 @@ class DiffractionObject:
     def on_d(self):
         return [self.all_arrays[:, 3], self.all_arrays[:, 0]]
 
-    def scale_to(self, target_diff_object, xtype=None, xvalue=None):
-        f"""
+    def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=0):
+        """
         returns a new diffraction object which is the current object but rescaled in y to the target
+
+        The y-value in the target at the closest specified x-value will be used as the factor to scale to.
+        The entire array is scaled by this factor so that one object places on top of the other at that point.
+        If multiple values of `q`, `tth`, or `d` are provided, the priority is `q` > `tth` > `d`.
+        If none are provided, the midpoint of the current object's `q`-array will be used.
 
         Parameters
         ----------
         target_diff_object: DiffractionObject
-          the diffraction object you want to scale the current one on to
-        xtype: string, optional.  Default is Q
-          the xtype, from {XQUANTITIES}, that you will specify a point from to scale to
-        xvalue: float. Default is the midpoint of the array
-          the y-value in the target at this x-value will be used as the factor to scale to.
-          The entire array is scaled be the factor that places on on top of the other at that point.
-          xvalue does not have to be in the x-array, the point closest to this point will be used for the scaling.
+            the diffraction object you want to scale the current one onto
+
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -390,14 +390,15 @@ def scale_to(self, target_diff_object, xtype=None, xvalue=None):
 
         data = self.on_xtype(xtype)
         target = target_diff_object.on_xtype(xtype)
+        if len(data[0]) == 0 or len(target[0]) == 0 or len(data[0]) != len(target[0]):
+            raise ValueError("I cannot scale two diffraction objects with empty or different lengths.")
```
yucongalicechen, line 394: @sbillinge I added this error message here but I'm not sure if this is appropriate. Does it make more sense that we find `xindex` for each diffraction object?

src/diffpy/utils/diffraction_objects.py
```diff
@@ -390,14 +390,15 @@ def scale_to(self, target_diff_object, xtype=None, xvalue=None):
...
         if xvalue is None:
             xvalue = data[0][0] + (data[0][-1] - data[0][0]) / 2.0
 
         xindex = (np.abs(data[0] - xvalue)).argmin()
         ytarget = target[1][xindex]
         yself = data[1][xindex]
```
yucongalicechen, line 400: Does it make more sense that we find an xindex for the original DO that we want to scale, and then another xindex for the target DO, and scale based on that?

yucongalicechen:
> @yucongalicechen I was explaining this functionality to colleagues in an email and the syntax that came most naturally to me was:
> 
> ```
> pyplot.plot(xpd_data.on_q[0], xpd_data.on_q[1]) 
> pyplot.plot(lab_data.on_q[0], lab_data.scaled_to(xpd_data, q=10, offset=5).on_q[1])
> pyplot.show()
> ```
> 
> which looks more intuitive to me (though less extensible) than `lab_data.scaled_to(xpd_data, xvalue=10, xtype=q, offset=5).on_q[1])`. Shall we do it this way? We will have to define as options `tth=None` and logic to figure out which one to use. It is a tradeoff, easier for the user vs. easier for us....
> 
> Also, please let's be in the habit of making the news early in the process to make it easier to see when tests are passing.

Yeah I also like this better!

yucongalicechen:
@sbillinge ready for another review

src/diffpy/utils/diffraction_objects.py
```diff
@@ -364,41 +364,43 @@ def on_tth(self):
...
+
+        offset : float, optional, default is 0
+            an offset to add to the scaled y-values
 
         Returns
         -------
         the rescaled DiffractionObject as a new object
-
         """
         scaled = deepcopy(self)
```
sbillinge, line 391: I think we can use @bobleesj nice copy method to do this now.  ACtually, it just does a deepcopy, but let's model syntax that we would like users to use..... so `scaled = self.copy()` or sthg like that?

sbillinge (commented):
few comments

sbillinge:
@yucongalicechen some of the comments might be out of date...i may have made them but not submitted.  Please just use common sense and ask if not sure.  If they don't make sense it is because we changed our desired api.

yucongalicechen:
@sbillinge ready for review

tests/test_diffraction_objects.py
```diff
@@ -223,6 +223,114 @@ def test_on_xtype_bad():
...
+            2 * np.pi,
+            0.1,
+            None,
+            None,
+            0,
+        ],
+        ["q", np.array([1, 2, 4, 6])],
+    ),
+    # UC4: different x-array lengths with approximate x-value match
+    (
```
yucongalicechen, line 282: A test example for scaling DOs with different array lengths. Here I think it makes more sense to scale them on q=61 (for self) & q=62 (for target).

src/diffpy/utils/diffraction_objects.py
```diff
@@ -364,40 +364,41 @@ def on_tth(self):
...
 
         Returns
         -------
         the rescaled DiffractionObject as a new object
-
         """
-        scaled = deepcopy(self)
-        if xtype is None:
-            xtype = "q"
+        scaled = self.copy()
```
yucongalicechen, line 391: changed deepcopy to self.copy

bobleesj:
@yucongalicechen this one just needs to be reviewed right?

sbillinge (commented):
Sorry, forgot to finish it up

tests/test_diffraction_objects.py
```diff
@@ -180,6 +180,144 @@ def test_init_invalid_xtype():
...
+
+@pytest.mark.parametrize("inputs", params_scale_to_bad)
+def test_scale_to_bad(inputs):
+    orig_diff_object = DiffractionObject(xarray=inputs[0], yarray=inputs[1], xtype=inputs[2], wavelength=inputs[3])
+    target_diff_object = DiffractionObject(
+        xarray=inputs[4], yarray=inputs[5], xtype=inputs[6], wavelength=inputs[7]
+    )
+    with pytest.raises(
+        ValueError, match="You can only specify one of 'q', 'tth', or 'd'. Please rerun specifying only one."
+    ):
```
yucongalicechen, line 317: added a test for error message

tests/test_diffraction_objects.py
```diff
@@ -180,6 +180,144 @@ def test_init_invalid_xtype():
...
+            2 * np.pi,
+            None,
+            60,
+            None,
+            0,
+        ],
+        # scaling factor is calculated at index = 5 for self and index = 6 for target
+        ["tth", np.array([1, 2, 3, 4, 5, 6, 10])],
+    ),
+    # UC5: user did not specify anything, use the midpoint of the current object's q-array
```
yucongalicechen, line 257: added a test for specifying nothing

yucongalicechen:
@sbillinge ready for review!

src/diffpy/utils/diffraction_objects.py
```diff
@@ -391,40 +391,45 @@ def on_tth(self):
...
-          the xtype, from {XQUANTITIES}, that you will specify a point from to scale to
-        xvalue: float. Default is the midpoint of the array
-          the y-value in the target at this x-value will be used as the factor to scale to.
-          The entire array is scaled be the factor that places on on top of the other at that point.
-          xvalue does not have to be in the x-array, the point closest to this point will be used for the scaling.
+            the diffraction object you want to scale the current one onto
+
+        q, tth, d : float, optional, default is the midpoint of the current object's `q`-array
+            the xvalue (in `q`, `tth`, or `d` space) to align the current and target objects
+
```
sbillinge, line 410: let's change the logic so that if no q, tth or d are specified then it crashes with an error.  I don't think anyone will ever want to use the default of midpoint and it can just cause problems because the code will work strangely when a value doesn't get passed by accident.  no magic....

src/diffpy/utils/diffraction_objects.py
```diff
@@ -364,40 +364,43 @@ def on_tth(self):
...
         if xvalue is None:
-            xvalue = data[0][0] + (data[0][-1] - data[0][0]) / 2.0
+            xvalue = (data[0][0] + data[0][-1]) / 2.0
 
-        xindex = (np.abs(data[0] - xvalue)).argmin()
-        ytarget = target[1][xindex]
-        yself = data[1][xindex]
-        scaled.on_tth[1] = data[1] * ytarget / yself
-        scaled.on_q[1] = data[1] * ytarget / yself
+        x_data, x_target = (np.abs(data[0] - xvalue)).argmin(), (np.abs(target[0] - xvalue)).argmin()
```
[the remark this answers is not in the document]
sbillinge: let's change the variable name to `xindex_data` to remind us that these are indices.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -391,40 +391,45 @@ def on_tth(self):
...
+        y_data, y_target = data[1][x_data], target[1][x_target]
```
sbillinge, line 431: I think this line makes things less readable.  I would put the `data[1][xindex_data]` (and so on) directly in the expression below.

sbillinge (commented):
please see inline

One other thought circling in my mind is to have the DO's carry around a kind of master array that kind of goes from 0 to q=45 (about the widest range we will ever encounter) with a fine grid and we interpolate the values onto there (linear or quadratic, whichever gives us sufficient accuracy.  We could use it for various things, but this would be one of them because we could compute the scaling for the arrays on this grid (but apply the scaling to the data on all the grids).  Don't make this change, but let's thin about it. The pro is it is cleaner and nicer and could be used for other things.  The con is that we have to carry around another largish array in memory.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -391,40 +391,44 @@ def on_tth(self):
...
         """
-        scaled = deepcopy(self)
-        if xtype is None:
-            xtype = "q"
+        scaled = self.copy()
+        count = sum([q is not None, tth is not None, d is not None])
+        if count != 1:
+            raise ValueError(
+                "You must specify exactly one of 'q', 'tth', or 'd'. Please rerun specifying only one."
+            )
```
yucongalicechen, line 422: add error message so that user has to specify one of q, tth, d arguments

src/diffpy/utils/diffraction_objects.py
```diff
@@ -391,40 +391,44 @@ def on_tth(self):
...
+        xvalue = q if xtype == "q" else tth if xtype == "tth" else d
 
-        xindex = (np.abs(data[0] - xvalue)).argmin()
-        ytarget = target[1][xindex]
-        yself = data[1][xindex]
-        scaled.on_tth[1] = data[1] * ytarget / yself
-        scaled.on_q[1] = data[1] * ytarget / yself
+        xindex_data = (np.abs(data[0] - xvalue)).argmin()
+        xindex_target = (np.abs(target[0] - xvalue)).argmin()
+        scaled._all_arrays[:, 0] = data[1] * target[1][xindex_target] / data[1][xindex_data] + offset
```
yucongalicechen, line 431: edit variable names

tests/test_diffraction_objects.py
```diff
@@ -180,6 +180,142 @@ def test_init_invalid_xtype():
...
+    scaled_diff_object = orig_diff_object.scale_to(
+        target_diff_object, q=inputs[8], tth=inputs[9], d=inputs[10], offset=inputs[11]
+    )
+    # Check the intensity data is same as expected
+    assert np.allclose(scaled_diff_object.on_xtype(expected[0])[1], expected[1])
+
+
+params_scale_to_bad = [
+    # UC1: user did not specify anything
+    (
```
yucongalicechen, line 275: add the bad test case for specifying nothing

yucongalicechen:
> please see inline
> 
> One other thought circling in my mind is to have the DO's carry around a kind of master array that kind of goes from 0 to q=45 (about the widest range we will ever encounter) with a fine grid and we interpolate the values onto there (linear or quadratic, whichever gives us sufficient accuracy. We could use it for various things, but this would be one of them because we could compute the scaling for the arrays on this grid (but apply the scaling to the data on all the grids). Don't make this change, but let's thin about it. The pro is it is cleaner and nicer and could be used for other things. The con is that we have to carry around another largish array in memory.

@sbillinge the new commit is ready for review. I like this idea and I think it will be very helpful for comparing absorption corrected curves. We can also have a function that allows user to specify which xarray to interpolate on?

tests/test_diffraction_objects.py
```diff
@@ -180,6 +180,142 @@ def test_init_invalid_xtype():
...
+@pytest.mark.parametrize("inputs", params_scale_to_bad)
+def test_scale_to_bad(inputs):
+    orig_diff_object = DiffractionObject(xarray=inputs[0], yarray=inputs[1], xtype=inputs[2], wavelength=inputs[3])
+    target_diff_object = DiffractionObject(
+        xarray=inputs[4], yarray=inputs[5], xtype=inputs[6], wavelength=inputs[7]
+    )
+    with pytest.raises(
+        ValueError, match="You must specify exactly one of 'q', 'tth', or 'd'. Please rerun specifying only one."
+    ):
+        orig_diff_object.scale_to(target_diff_object, q=inputs[8], tth=inputs[9], d=inputs[10], offset=inputs[11])
```
bobleesj, line 316: having inputs up to ,etc `inputs[10]` does not appear scalable and I found this was very hard to read and maintain in diffpy.snmf which I had to refactor: https://github.com/diffpy/diffpy.snmf/pull/120/files#diff-1bd6af744434d75c63490430b955f577f60277dfe95e9ad716e3f808a2ed9d48L85-L87

Discussion here:
https://github.com/diffpy/diffpy.utils/pull/225#discussion_r1884266972

bobleesj: One way to resolve this future nightware could be having reusable instances of `DiffractionObject` defined under `conftest.py` with specific UC cases. Then, we import these instances through the parameters in each test func. Thoughts?

sbillinge:
> > please see inline
> > One other thought circling in my mind is to have the DO's carry around a kind of master array that kind of goes from 0 to q=45 (about the widest range we will ever encounter) with a fine grid and we interpolate the values onto there (linear or quadratic, whichever gives us sufficient accuracy. We could use it for various things, but this would be one of them because we could compute the scaling for the arrays on this grid (but apply the scaling to the data on all the grids). Don't make this change, but let's thin about it. The pro is it is cleaner and nicer and could be used for other things. The con is that we have to carry around another largish array in memory.
> 
> @sbillinge the new commit is ready for review. I like this idea and I think it will be very helpful for comparing absorption corrected curves. We can also have a function that allows user to specify which xarray to interpolate on?

yes, that had occured to me too.  I think that was where my functions came from for building arrays using arange etc..  But for most flexibility for hte least code, I would suggest we just allow the user to pass in a master array.  We can force them to do it on "q".   If they want to specify a tth value to scale to we will have to convert.

@yucongalicechen will you put these on new issues?  IN any case, let's put them on milestone 3.7.  

Are you finished working on this PR and it can be merged?

tests/test_diffraction_objects.py
sbillinge: yes, I agree in this case, this  would be helpful.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -391,40 +391,44 @@ def on_tth(self):
...
-        xtype: string, optional.  Default is Q
-          the xtype, from {XQUANTITIES}, that you will specify a point from to scale to
-        xvalue: float. Default is the midpoint of the array
-          the y-value in the target at this x-value will be used as the factor to scale to.
-          The entire array is scaled be the factor that places on on top of the other at that point.
-          xvalue does not have to be in the x-array, the point closest to this point will be used for the scaling.
+            the diffraction object you want to scale the current one onto
+
+        q, tth, d : float, optional, must specify exactly one of them
+            the xvalue (in `q`, `tth`, or `d` space) to align the current and target objects
```
sbillinge, line 408: "The value of the x-array where you want the curves to line up vertically.  Specify a value on one of the allowed grids, `q`, `tth`, or `d`), e.g., q=10."

src/diffpy/utils/diffraction_objects.py
```diff
@@ -391,40 +391,44 @@ def on_tth(self):
...
-        if xtype is None:
-            xtype = "q"
+        scaled = self.copy()
+        count = sum([q is not None, tth is not None, d is not None])
+        if count != 1:
+            raise ValueError(
+                "You must specify exactly one of 'q', 'tth', or 'd'. Please rerun specifying only one."
+            )
+
+        xtype = "q" if q is not None else "tth" if tth is not None else "d" if d is not None else "q"
```
sbillinge, line 424: can we drop the last "else "q""? given our validation above?

src/diffpy/utils/diffraction_objects.py
```diff
@@ -391,40 +391,44 @@ def on_tth(self):
...
+        data, target = self.on_xtype(xtype), target_diff_object.on_xtype(xtype)
```
sbillinge, line 425: split to two lines for greater readability

tests/test_diffraction_objects.py
sbillinge: btw, to make it more readable we could also pass the inputs as a dict so it would read `input["wavelenght"]` instead of `input[0]`.  The intent of the former is much clearer.

sbillinge (commented):
please see inline.

yucongalicechen:
thanks for the suggestions @sbillinge @bobleesj . I've edited the docstring and tests to make the codes more readable. Also created a new issue #230 for the comment above. This PR is ready to be reviewed again.

PR #214, opened 2024-12-12
Remove unused test util functions for comparing two dicts, use `__eq__` to compare DiffractionObjects

bobleesj:


tests/test_diffraction_objects.py
```diff
@@ -193,14 +156,9 @@ def dicts_equal(dict1, dict2):
...
 @pytest.mark.parametrize("inputs1, inputs2, expected", params)
 def test_diffraction_objects_equality(inputs1, inputs2, expected):
-    diffraction_object1 = DiffractionObject(**inputs1)
-    diffraction_object2 = DiffractionObject(**inputs2)
-    # diffraction_object1_attributes = [key for key in diffraction_object1.__dict__ if not key.startswith("_")]
-    # for i, attribute in enumerate(diffraction_object1_attributes):
-    #     setattr(diffraction_object1, attribute, inputs1[i])
-    #     setattr(diffraction_object2, attribute, inputs2[i])
-    print(dicts_equal(diffraction_object1.__dict__, diffraction_object2.__dict__), expected)
-    assert dicts_equal(diffraction_object1.__dict__, diffraction_object2.__dict__) == expected
```
bobleesj, line 203: Since we will be introducing an unique `uuid` via `_id`, this test will fail. But rather we use `__eq__` method.

tests/test_diffraction_objects.py
```diff
@@ -8,43 +8,6 @@
 
 from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
 
-
-def compare_dicts(dict1, dict2):
```
bobleesj, line 12: These functions are no longer used in our test code - We. may use the `DeepDiff` library instead to compare two dict as implemented in our previous PR:

```python
    actual_do = DiffractionObject(**inputs)
    diff = DeepDiff(actual_do.__dict__, expected, ignore_order=True, significant_digits=13)
    assert diff == {}
```

bobleesj (commented):
While I was waiting for other PRs to be merged, I was reading the code and found these potential areas of improvement. Please let me know! @sbillinge

tests/test_diffraction_objects.py
sbillinge: @bobleesj yes, this is much cleaner.  A sure sign that our code and API is improving, don't you think?

sbillinge (commented):
yes, thanks for this @bobleesj

tests/test_diffraction_objects.py
bobleesj: absolutely

Issue #213, opened 2024-12-10
add uuids to DiffractionObjects


sbillinge:

bobleesj:
@sbillinge 

```python
import uuid

class DiffractionObjects:
    def __init__(self, xarray, yarray, xtype, ...):
        ....
        self._uuid = uuid.uuid4()

    @property
    def uuid(self):
        """Get the UUID of the scattering quantity."""
        return self._uuid
```

Something like this? if it is, happy to do this for 3.6.0

bobleesj:
Checked the slack msg.

PR #216, opened 2024-12-12
Add gettable `id` property to `DiffractionObject`

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/213

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -30,7 +30,7 @@ def _xtype_wmsg(xtype):
 def _setter_wmsg(attribute):
     return (
         f"Direct modification of attribute '{attribute}' is not allowed. "
-        f"Please use 'input_data' to modify '{attribute}'.",
+        f"Please use 'insert_scattering_quantity' to modify '{attribute}'.",
     )
 
 
@@ -55,7 +55,7 @@ class DiffractionObject:
             yarray = np.empty(0)
 
         self._id = uuid.uuid4()
-        self.input_data(xarray, yarray, xtype)
+        self.insert_scattering_quantity(xarray, yarray, xtype)
 
     def __eq__(self, other):
         if not isinstance(other, DiffractionObject):
@@ -327,7 +327,7 @@ class DiffractionObject:
         self.dmin = np.nanmin(self._all_arrays[:, 3], initial=np.inf)
         self.dmax = np.nanmax(self._all_arrays[:, 3], initial=0.0)
 
-    def input_data(
+    def insert_scattering_quantity(
         self,
         xarray,
         yarray,
@@ -361,7 +361,7 @@ class DiffractionObject:
         if len(xarray) != len(yarray):
             raise ValueError(
                 "'xarray' and 'yarray' must have the same length. "
-                "Please re-initialize 'DiffractionObject' or re-run the method 'input_data' "
+                "Please re-initialize 'DiffractionObject' or re-run the method 'insert_scattering_quantity' "
                 "with 'xarray' and 'yarray' of identical length."
             )
 
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -1,4 +1,5 @@
 import datetime
+import uuid
 import warnings
 from copy import deepcopy
 
@@ -29,7 +30,7 @@ def _xtype_wmsg(xtype):
 def _setter_wmsg(attribute):
     return (
         f"Direct modification of attribute '{attribute}' is not allowed. "
-        f"Please use 'insert_scattering_quantity' to modify '{attribute}'.",
+        f"Please use 'input_data' to modify '{attribute}'.",
     )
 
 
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -205,6 +207,14 @@ def input_xtype(self):
     def input_xtype(self, _):
         raise AttributeError(_setter_wmsg("input_xtype"))
 
+    @property
+    def id(self):
+        return self._id
+
+    @id.setter
+    def id(self, _):
+        raise AttributeError(_setter_wmsg("id"))
```
bobleesj, line 216: `id` not settable

tests/test_diffraction_objects.py
```diff
@@ -335,8 +337,8 @@ def test_dump(tmp_path, mocker):
 
 @pytest.mark.parametrize("inputs, expected", tc_params)
 def test_constructor(inputs, expected):
-    actual_do = DiffractionObject(**inputs)
-    diff = DeepDiff(actual_do.__dict__, expected, ignore_order=True, significant_digits=13)
+    actual = DiffractionObject(**inputs).__dict__
+    diff = DeepDiff(actual, expected, ignore_order=True, significant_digits=13, exclude_paths="root['_id']")
```
bobleesj, line 341: for testing `constructor`, `_id` is ignored since it's hard to dynamically compare its uuid value.

bobleesj: is that ok? not sure if there is a better way for this.

tests/test_diffraction_objects.py
```diff
@@ -369,6 +371,23 @@ def test_all_array_setter():
         actual_do.all_arrays = np.empty((4, 4))
 
 
+def test_id_getter():
+    do = DiffractionObject()
+    assert hasattr(do, "id")
+    assert isinstance(do.id, UUID)
+    assert len(str(do.id)) == 36
```
bobleesj, line 378: for testing `id`, check the the id length,  its type, its existence.

tests/test_diffraction_objects.py
```diff
@@ -384,7 +403,7 @@ def test_input_xtype_getter():
     assert do.input_xtype == "tth"
 
 
-def test_input_xtype_setter():
+def test_input_xtype_setter_error():
```
bobleesj, line 406: just enhancing the function name for clarify

bobleesj (commented):
@sbillinge ready for review with in-line comments added for your feedback.

tests/test_diffraction_objects.py
sbillinge: In the past I have mocked that attribute.  This seems less work.  If we have tests that test the id setting and getting, we can safely ignore it in other tests that are testing other things I think

tests/test_diffraction_objects.py
sbillinge: Here I would probably mock it

sbillinge (commented):
This all looks good to me.  I am happy to merge it but please see comments and let me know

tests/test_diffraction_objects.py
```diff
@@ -369,6 +371,29 @@ def test_all_array_setter():
...
 
+def test_id_getter():
+    do = DiffractionObject()
+    assert hasattr(do, "id")
+    assert isinstance(do.id, UUID)
+    assert len(str(do.id)) == 36
+
+
+def test_id_getter_with_mock(mocker):
+    mocker.patch.object(DiffractionObject, "id", new_callable=lambda: UUID("d67b19c6-3016-439f-81f7-cf20a04bee87"))
```
bobleesj, line 382: Since `id` is used with `@property`, the following doesn't apply:

```
mocker.patch.object(DiffractionObject, 'id', return_value="d67b19c6-3016-439f-81f7-cf20a04bee87")
  ```

Hence we use the `new_callable` parameter.

tests/test_diffraction_objects.py
bobleesj: Noted - yes is setting/getting provided below.

tests/test_diffraction_objects.py
bobleesj: Yes, I've maintained this test since we still want to have a getter test for `id` but also added a mock right below.

bobleesj:
@sbillinge ready for review - codecov is causing an issue although I re-ran the CI with an empty commit. 

Screenshot 2024-12-12 at 5 33 49 PM

tests/test_diffraction_objects.py
sbillinge: I see, nicely done.  I didn't know about that.

sbillinge (approved):

sbillinge:
@bobleesj @yucongalicechen I merged this but now I am having second thoughts. Do we want the id associated with the data rather than the DO itself? In that case we should mint it in `load_data` and not in the constructor.

More discussion: this would go away if we don't allow an empty instance where data are loaded later.  I started leaning that way because I couldn't think of a UC where we would really need that, though one would be if someone wants to load directly from a file to the DO.  We could also allow that as an option in the constructor too.  Can you have a think about it?

If we do mint the id on the load data function, all of our setters will have to have it insterted too.

Summary: maybe the cleanest is to mint it in the constructor as we do now but remove the ability to load data outside of the constructor (and make that method private).  But aloe an option to instantiate from file by giving a file path instead of a set of arrays.

What do you think? I think I like that...

bobleesj:
@sbillinge  

since this PR has been merged, I created an issue https://github.com/diffpy/diffpy.utils/issues/224 to continue our discussions

PR #217, opened 2024-12-12
Add doc build GitHub CI yml file

bobleesj:
Back in the days, we used to have this combined within the release script but we've then decided to have it as a separate .yml file so that we can run via workflow dispatch easily.


bobleesj:
Ha, codecov is working

bobleesj:
@sbillinge ready for review and merge

Issue #195, opened 2024-12-04
write a DiffractionObject serializer/deserializer


sbillinge:
### Problem
It would be great if we could dump and load DiffractionObjects as DiffractionObjects

### Proposed solution
either use json, but maybe better, figure out how to do it with an hdf5 file to preserve the numpy arrays correctly, but also all the metadata.

Maybe save the file with a `dfpydo` extension or sthg like that.  This would allow DiffractionObjects to be shared more easily, with all their metadata.

bobleesj:
Good comments by @sbillinge from https://github.com/diffpy/diffpy.utils/pull/222

> I think we probably want code that will dump to different file formats and read from different formats. These are dumpers and loaders. If we serialize and deserialize things I think of it as a kind of (safe) round trip. the unsafe things is just to pickle and then unpickle but we need to be more careful now. But sinnce we want people using diffraction objects, the serializers that most interesting to me would be ones that serialize and deserialize complete diffraction objects.

> We definitely want the dumpers and loaders and a pure serializer. We can put it off to 3.7 though to keep momentum. The serializer will change the API though because I would like to give as an option a file path, or probably better a stream so we can have different backends, and have it load the DO from there. So the serializer would be nice now I guess?

> Dumpers and loaders to group standards file formats as we have are a bit specific, but probably worth having since the large community is using these file formats

We could re-visit this once we have `3.6.0` milestone delivered?

bobleesj:
Recording today's in-person discussion with @sbillinge 

- **Serialization** takes a DiffractionObject (DO) and outputs it to a file (.txt, .json, etc.), which can be imported into databases and shared via email with colleagues. There is no loss of information.
- **Deserialization** uses the the serialized file to instantiate a DO object. E

How is this different from using a loader/dumper?

- We want serialized DOs to be usable by other existing software, e.g., GSAS.
- A loader/dumper saves a file in a format that the existing software can read.
- There is some information loss, but it's a much better solution than not being able to import a DO into these packages or waiting for these packages to support our format, which may not happen soon.

bobleesj:
Note to myself: two UC provided by @sbillinge in using DiffractionObject with empty arrays for xarray and yarray: https://github.com/diffpy/diffpy.utils/issues/195

Issue #27, opened 2023-08-12
separate data validation and serialization in serialize.py


sbillinge:
the serialize and deserialize functions contain extensive code for validating that the data is of the right form.  This functionaliity would useful as a separate function in a module called `diffpy.utils.validators.py` where I could easily check that all my multi-dimensional data is the right shape and so on, regardless of whether I got it from loadData or not.

This issue is to pull the validation code out of serialize into validate functions that are available separaterly and then change the serialize code so it calls the validate functions to give the same functionality.

bobleesj:
Per our discussion in https://github.com/diffpy/diffpy.utils/pull/222, we might want to refactor the following code and while doing so, we can separate validation functions as needed. 

```python

def serialize_data(
    filename,
    hdata: dict,
    data_table,
    dt_colnames=None,
    show_path=True,
    serial_file=None,
):
    """Serialize file data into a dictionary. Can also save dictionary into a serial language file. Dictionary is
    formatted as {filename: data}.

    Requires hdata and data_table (can be generated by loadData).

    Parameters
    ----------
    filename
        Name of the file whose data is being serialized.
    hdata: dict
        File metadata (generally related to data table).
    data_table: list or ndarray
        Data table.
    dt_colnames: list
        Names of each column in data_table. Every name in data_table_cols will be put into the Dictionary
        as a key with a value of that column in data_table (stored as a List). Put None for columns
        without names. If dt_cols has less non-None entries than columns in data_table,
        the pair {'data table': data_table} will be put in the dictionary.
        (Default None: only entry {'data table': data_table} will be added to dictionary.)
    show_path: bool
        include a path element in the database entry (default True). If 'path' is not included in hddata,
         extract path from filename.
    serial_file
        Serial language file to dump dictionary into. If None (default), no dumping will occur.

    Returns
    -------
    dict:
        Returns the dictionary loaded from/into the updated database file.
    """

    # compile data_table and hddata together
    data = {}

    # handle getting name of file for variety of filename types
    abs_path = pathlib.Path(filename).resolve()
    # add path to start of data if requested
    if show_path and "path" not in hdata.keys():
        data.update({"path": abs_path.as_posix()})
    # title the entry with name of file (taken from end of path)
    title = abs_path.name

    # first add data in hddata dict
    data.update(hdata)

    # second add named columns in dt_cols
    # performed second to prioritize overwriting hdata entries with data_table column entries
    named_columns = 0  # initial value
[...]

PR #222, opened 2024-12-13
Refactor serialize_data with better param names and logic

bobleesj:


tests/test_serialization.py
```diff
@@ -17,30 +17,29 @@ def test_load_multiple(tmp_path, datafile):
...
     generated_data = None
+
     for headerfile in tlm_list:
         # gather data using loadData
         hdata = loadData(headerfile, headers=True)
         data_table = loadData(headerfile)
 
         # check path extraction
-        generated_data = serialize_data(headerfile, hdata, data_table, dt_colnames=["r", "gr"], show_path=True)
+        generated_data = serialize_data(headerfile, hdata, data_table, dt_col_names=["r", "gr"], show_path=True)
```
bobleesj, line 27: `df_col_names` and `df_colnames` are used across, so sticking with `dt_col_names`

src/diffpy/utils/parsers/serialization.py
```diff
@@ -64,64 +64,69 @@ def serialize_data(
...
+    # Prioritize overwriting hdata entries with data_table column entries
+    col_counter = 0
+
+    # Get a list of column counts in each entry in data table
+    dt_col_counts = [len(row) for row in data_table]
+    dt_max_col_count = max(dt_col_counts)
+
+    if dt_col_names is not None:
+        dt_col_names_count = len(dt_col_names)
+        if dt_max_col_count < dt_col_names_count:  # assume numpy.loadtxt gives non-irregular array
```
bobleesj, line 89: Compare the number of `col_names` values provided as an input for this function against the number of col determined from the data table.

bobleesj (commented):
@sbillinge ready for review. I was looking into this issue: https://github.com/diffpy/diffpy.utils/issues/27

Before addressing this issue, I had to understand the code and thought I could help refactor this.

src/diffpy/utils/parsers/serialization.py
```diff
@@ -40,13 +40,13 @@ def serialize_data(
 
     Parameters
     ----------
-    filename
-        Name of the file whose data is being serialized.
+    filepath
```
bobleesj, line 43:
```python
    for headerfile in tlm_list:
        # gather data using loadData
        hdata = loadData(headerfile, headers=True)
        data_table = loadData(headerfile)

        # check path extraction
        generated_data = serialize_data(headerfile...)
```

Here `headerfile` is used for `serialize_data(filepath, ...)`. `headerfile` prints `/Users/macbook/Downloads/dev/diffpy.utils/tests/testdata/dbload/e1.gr`

bobleesj:
> @sbillinge ready for review. I was looking into this issue: #27
> 
> Before addressing this issue, I had to understand the code and thought I could help refactor this.

Ready for review

src/diffpy/utils/parsers/serialization.py
```diff
@@ -40,13 +40,13 @@ def serialize_data(
...
+        The file path whose data is being serialized.
     hdata: dict
-        File metadata (generally related to data table).
+        The file metadata (generally related to data table).
```
sbillinge, line 46: is there a reason we are not calling this variable `metadata`?  or `header_data`.   calling it `hdata` I am having to carry around too many pointers in my brain that I have to cognitively resolve as I read.

src/diffpy/utils/parsers/serialization.py
```diff
@@ -40,13 +40,13 @@ def serialize_data(
...
     data_table: list or ndarray
-        Data table.
-    dt_colnames: list
+        The data table.
```
sbillinge, line 48: what is a data table?  A table with data on it?  I think this is the something like the column data read from the file or something.

sbillinge (commented):
I started to review this, but @bobleesj like you I am a bit confused what is going on here right from the get-go.  The function is called `serialize_data` but it seems to be reading  data-file.   This seems closer to deserializing data, not serializing it.  We may want to get @Sparks29032 in on the convo here, I think he may have written it.

I think we probably want code that will dump to different file formats and read from different formats.  These are dumpers and loaders.  If we serialize and deserialize things I think of it as a kind of (safe) round trip.  the unsafe things is just to pickle and then unpickle but we need to be more careful now.  But sinnce we want people using diffraction objects, the serializers that most interesting to me would be ones that serialize and deserialize complete diffraction objects.

bobleesj:
@Sparks29032 when you have time this Sat,  is the `serialize_data` method used in any of the diffpy packages? Wondering the motivation and how it's used in practice.

Sparks29032:
I think this was originally added for the `pdfitc` rest-api project. I don't believe it is implemented in any packages.

These are dumpers and loaders, but called serialization/deserialization as we are moving from serial files to python-readable information.

This was coded a while back, so I don't remember all the details, but I believe these were coded specifically to dump and load PDF data (hence a data table [list type] + header information [dictionary type]) into serial file formats like `.json`.

Moving forward, it might be good to standardize so only dictionaries can be passed into these functions, and the dumpers and loaders can use specific file-format libraries given the extension of the file.

src/diffpy/utils/parsers/serialization.py
Sparks29032: The hdata came from loadData's naming of hdata for header data information. Good to change to metadata.

src/diffpy/utils/parsers/serialization.py
Sparks29032: These serialization functions were coded in tandem with loadData, so the inputs of these match outputs of loadData. This assumes files structured like .chi/.gr/etc. files.

bobleesj:
Thanks @Sparks29032 

> I think this was originally added for the pdfitc rest-api project. I don't believe it is implemented in any packages.

To confirm, this function is used by `pdfitc`? Wasn't sure if the source code for pdffitc is publicly available. 

> load PDF data (hence a data table [list type] + header information

Yes, there are test cases for loading `.gr` files  and turning these into a single `.json` file.

@sbillinge As the next step, since we have an issues to write a DiffractionObject serializer/deserializer (https://github.com/diffpy/diffpy.utils/issues/195) in `3.7.0`, should we close the following issue https://github.com/diffpy/diffpy.utils/issues/27 that this PR is trying to slowly address?

Sparks29032:
@bobleesj It is not used by pdfitc, don't worry. It was experimental.

sbillinge:
We definitely want the dumpers and loaders and a pure serializer.  We can put it off to 3.7 though to keep momentum. The serializer will change the API though because I would like to give as an option a file path, or probably better a stream so we can have different backends, and have it load the DO from there. So the serializer would be nice now I guess?

sbillinge:
Dumpers and loaders to group standards file formats as we have are a bit specific, but probably worth having since the large community is using these file formats

bobleesj:
Since this is a PR, will copy and paste the comments by @sbillinge in the existing issue here https://github.com/diffpy/diffpy.utils/issues/195.

bobleesj:
Closing - we will probably make a new PR while when we are dealing with `3.7.0`. For now, this PR isn't needed for `3.6.0` release.

PR #225, opened 2024-12-13
1) Capture no wavelength `UserWarning` for `test_q_to_thh` method 2) discuss passing variables to `@pytest.mark.parametrize`

bobleesj:
Here, the goal is to discuss how to write good tests using `@pytest.mark.parametrize` and also test warning messages.

### DIFF
--- a/src/diffpy/utils/transforms.py
+++ b/src/diffpy/utils/transforms.py
@@ -4,12 +4,10 @@ from copy import copy
 import numpy as np
 
 wavelength_warning_emsg = (
-    "INFO: no wavelength has been specified. You can continue "
-    "to use the DiffractionObject but some of its powerful features "
-    "will not be available. To specify a wavelength, set "
-    "diffraction_object.wavelength = [number], "
-    "where diffraction_object is the variable name of you Diffraction Object, "
-    "and number is the wavelength in angstroms."
+    "No wavelength has been specified. You can continue to use the DiffractionObject, but "
+    "some of its powerful features will not be available. "
+    "To specify a wavelength, if you have do = DiffractionObject(xarray, yarray, 'tth'), "
+    "you may set do.wavelength = 1.54 with the unit in angstroms."
 )
 invalid_tth_emsg = "Two theta exceeds 180 degrees. Please check the input values for errors."
 invalid_q_or_d_or_wavelength_emsg = (
--- a/src/diffpy/utils/transforms.py
+++ b/src/diffpy/utils/transforms.py
@@ -7,7 +7,7 @@ wavelength_warning_emsg = (
     "No wavelength has been specified. You can continue to use the DiffractionObject, but "
     "some of its powerful features will not be available. "
     "To specify a wavelength, if you have do = DiffractionObject(xarray, yarray, 'tth'), "
-    "you may set do.wavelength = 1.54 with the unit in angstroms."
+    "you may set do.wavelength = 1.54 for a wavelength of 1.54 angstroms."
 )
 invalid_tth_emsg = "Two theta exceeds 180 degrees. Please check the input values for errors."
 invalid_q_or_d_or_wavelength_emsg = (
--- a/tests/test_diffraction_objects.py
+++ b/tests/test_diffraction_objects.py
@@ -164,11 +164,11 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
 
 
 def test_on_xtype():
-    test = DiffractionObject(wavelength=2 * np.pi, xarray=np.array([30, 60]), yarray=np.array([1, 2]), xtype="tth")
-    assert np.allclose(test.on_xtype("tth"), [np.array([30, 60]), np.array([1, 2])])
-    assert np.allclose(test.on_xtype("2theta"), [np.array([30, 60]), np.array([1, 2])])
-    assert np.allclose(test.on_xtype("q"), [np.array([0.51764, 1]), np.array([1, 2])])
-    assert np.allclose(test.on_xtype("d"), [np.array([12.13818, 6.28319]), np.array([1, 2])])
+    do = DiffractionObject(wavelength=2 * np.pi, xarray=np.array([30, 60]), yarray=np.array([1, 2]), xtype="tth")
+    assert np.allclose(do.on_xtype("tth"), [np.array([30, 60]), np.array([1, 2])])
+    assert np.allclose(do.on_xtype("2theta"), [np.array([30, 60]), np.array([1, 2])])
+    assert np.allclose(do.on_xtype("q"), [np.array([0.51764, 1]), np.array([1, 2])])
+    assert np.allclose(do.on_xtype("d"), [np.array([12.13818, 6.28319]), np.array([1, 2])])
 
 
 def test_init_invalid_xtype():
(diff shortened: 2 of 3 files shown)


tests/test_transforms.py
```diff
@@ -5,24 +5,31 @@
...
-    ([4 * np.pi, np.array([0, 1 / np.sqrt(2), 1.0])], np.array([0, 90.0, 180.0])),
+    (4 * np.pi, np.array([0, 1 / np.sqrt(2), 1.0]), np.array([0, 90.0, 180.0])),
 ]
 
 
-@pytest.mark.parametrize("inputs, expected", params_q_to_tth)
-def test_q_to_tth(inputs, expected):
-    actual = q_to_tth(inputs[1], inputs[0])
-    assert np.allclose(expected, actual)
+@pytest.mark.parametrize("wavelength, q, expected_tth", params_q_to_tth)
```
bobleesj, line 23: @sbillinge @yucongalicechen 

Before addressing other pytest warnings, I just discuss our group's way for writing tests. 

Here, instead of passing `input[1]` or `input[0]`, it's actually difficult for me (new user) to understand what these variables refer to. 

Moving forward, wouldn't it easier to specify them as `wavelength` and `q` next to  `mark.parametrize` instead? so that we can directly use these methods and the tests are much. more readable.

tests/test_transforms.py
```diff
@@ -5,24 +5,31 @@
...
+def test_q_to_tth(wavelength, q, expected_tth):
+
+    if wavelength is None:
+        with pytest.warns(UserWarning, match="INFO: no wavelength has been specified"):
```
bobleesj, line 27: Here, I am not writing the entire warning msg: 

```
UserWarning: INFO: no wavelength has been specified. You can continue to use the DiffractionObject but some of its powerful features will not be available. To specify a wavelength, set diffraction_object.wavelength = [number], where diffraction_object is the variable name of you Diffraction Object, and number is the wavelength in angstroms.
```

assuming we have a specific test func for this warning msg.

bobleesj:
@sbillinge ready for review - cc'ing @yucongalicechen for our discussions moving forward.

tests/test_transforms.py
```diff
@@ -5,24 +5,31 @@
 
 params_q_to_tth = [
     # UC1: Empty q values, no wavelength, return empty arrays
-    ([None, np.empty((0))], np.empty((0))),
-    # UC2: Empty q values, wavelength specified, return empty arrays
-    ([4 * np.pi, np.empty((0))], np.empty(0)),
+    (None, np.empty((0)), np.empty((0))),
+    # # UC2: Empty q values, wavelength specified, return empty arrays
```
bobleesj, line 9: Remove extra `# 0`

tests/test_transforms.py
sbillinge: I like the user warning written out because the test defines the behavior.  So the workflow is to write the test and review it until we like it.  Then write the code.  Here, it slipped by but we don't want "INFO" here. It is either a WARNING or an INFO, it can't be both.   Also, I think when we are describing code it is better to write the code than describe it in words, so say "....   To speciffy a wavelength, if `my_object = DiffractionObject(xarray, yarray, "tth") then you can type    `my_object.wavelength = 1.54` if the wavelength is 1.54 angstroms."  Something like that.

If test the same message in more than one test we can assign it to a variable.  But it is better if it is closer to the test and we are not importing it orsomething.

tests/test_transforms.py
sbillinge: yes, this makes sense.  I agree.   Since tests test behavior, the easier it is to see the behvaior, the better.  Sometimes it can be a bit hard with the `parameterize` decorator, but this is a good way to make it easier.  I agree.

tests/test_transforms.py
bobleesj: Yes, done.

tests/test_transforms.py
```diff
@@ -1,81 +1,97 @@
...
-def test_q_to_tth(inputs, expected):
-    actual = q_to_tth(inputs[1], inputs[0])
-    assert np.allclose(expected, actual)
+@pytest.mark.parametrize("wavelength, q, expected_tth", params_q_to_tth)
+def test_q_to_tth(wavelength, q, expected_tth):
+
+    wavelength_warning_emsg = (
+        "No wavelength has been specified. You can continue to use the DiffractionObject, but "
+        "some of its powerful features will not be available. "
+        "To specify a wavelength, if you have do = DiffractionObject(xarray, yarray, 'tth'), "
```
bobleesj, line 31: As suggested, provide a specific instruction how to set `wavelength`

tests/test_transforms.py
```diff
@@ -1,81 +1,97 @@
...
+        "Please check these values and re-instantiate the DiffractionObject with correct values.",
     ),
 ]
 
 
-@pytest.mark.parametrize("inputs, expected", params_q_to_tth_bad)
-def test_q_to_tth_bad(inputs, expected):
-    with pytest.raises(expected[0], match=expected[1]):
-        q_to_tth(inputs[1], inputs[0])
+@pytest.mark.parametrize("q, wavelength, expected_error_type, expected_error_msg", params_q_to_tth_bad)
```
bobleesj, line 63: I think it's much easier to read.. expected.. sth sth sth..

bobleesj (commented):
@sbillinge ready for review - there are still warnings but prob it would be best to create a couple of PRs by theme.

tests/test_diffraction_objects.py
```diff
@@ -164,11 +164,11 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
 
 
 def test_on_xtype():
```
bobleesj, line 166: I think we could use `@pytest.mark.parametrize` for this test function?

tests/test_diffraction_objects.py
```diff
@@ -164,11 +164,11 @@ def test_diffraction_objects_equality(inputs1, inputs2, expected):
...
-    test = DiffractionObject(wavelength=2 * np.pi, xarray=np.array([30, 60]), yarray=np.array([1, 2]), xtype="tth")
-    assert np.allclose(test.on_xtype("tth"), [np.array([30, 60]), np.array([1, 2])])
```
bobleesj, line 168: using the instance `do` seems like a better wording here?

src/diffpy/utils/transforms.py
```diff
@@ -4,12 +4,10 @@
...
-    "INFO: no wavelength has been specified. You can continue "
-    "to use the DiffractionObject but some of its powerful features "
-    "will not be available. To specify a wavelength, set "
-    "diffraction_object.wavelength = [number], "
-    "where diffraction_object is the variable name of you Diffraction Object, "
-    "and number is the wavelength in angstroms."
+    "No wavelength has been specified. You can continue to use the DiffractionObject, but "
+    "some of its powerful features will not be available. "
+    "To specify a wavelength, if you have do = DiffractionObject(xarray, yarray, 'tth'), "
+    "you may set do.wavelength = 1.54 with the unit in angstroms."
```
sbillinge, line 10: Maybe it is clearer to say <number> than 1.54 here, it add "for a wavelength of 1.54 angstroms".  Which would someone like Fillipo find clearer?

tests/test_diffraction_objects.py
sbillinge: Yes, test is now such a good choice.

tests/test_diffraction_objects.py
sbillinge: Yes, let's refactor

sbillinge: Actually, this could be a good use of standard DOs in contftest if we also have the properly computed arrays that we can check against with well thought it names.

sbillinge: E.g, sthg like `test_do` that shortens to to and instead of 2*np.pi we have`tdo_lambda` and so on. We just want it more readable everywhere so the intent comes through better and it is easier to maintain

tests/test_transforms.py
```diff
@@ -23,8 +25,15 @@
 @pytest.mark.parametrize("wavelength, q, expected_tth", params_q_to_tth)
 def test_q_to_tth(wavelength, q, expected_tth):
 
+    wavelength_warning_emsg = (
+        "No wavelength has been specified. You can continue to use the DiffractionObject, but "
+        "some of its powerful features will not be available. "
+        "To specify a wavelength, if you have do = DiffractionObject(xarray, yarray, 'tth'), "
+        "you may set do.wavelength = 1.54 with the unit in angstroms."
```
sbillinge, line 32: Please see the small tweak above

sbillinge (commented):
These improvements are excellent! Thanks @bobleesj !  I left a couple of small tweaks and comments on your suggestions to make it even better.  I think done of these lessons can carry over to better group standards in general so it is not wasted effort to get this right 👍

tests/test_diffraction_objects.py
bobleesj: https://github.com/diffpy/diffpy.utils/issues/229

Yup, created an issue. I will handle this.

bobleesj:
> These improvements are excellent! Thanks @bobleesj ! I left a couple of small tweaks and comments on your suggestions to make it even better. I think done of these lessons can carry over to better group standards in general so it is not wasted effort to get this right 👍

Yes, I am recording these lessons. I will sync with the GitLab page: https://gitlab.thebillingegroup.com/resources/group-wiki/-/wikis/Lessons-learned-in-Billinge-Group

src/diffpy/utils/transforms.py
```diff
@@ -4,12 +4,10 @@
...
-    "INFO: no wavelength has been specified. You can continue "
-    "to use the DiffractionObject but some of its powerful features "
-    "will not be available. To specify a wavelength, set "
-    "diffraction_object.wavelength = [number], "
-    "where diffraction_object is the variable name of you Diffraction Object, "
-    "and number is the wavelength in angstroms."
+    "No wavelength has been specified. You can continue to use the DiffractionObject, but "
+    "some of its powerful features will not be available. "
+    "To specify a wavelength, if you have do = DiffractionObject(xarray, yarray, 'tth'), "
+    "you may set do.wavelength = 1.54 for a wavelength of 1.54 angstroms."
```
bobleesj, line 10: `you may set do.wavelength = 1.54 for a wavelength of 1.54 angstroms` is clearer than what I had earlier.

bobleesj (commented):
@sbillinge ready for review - the only modification was done on `wavelength_warning_emsg` as shown in the latest commit.

Issue #227, opened 2024-12-14
Do not allow an empty instance of `DiffractionObject` - init requires `xarray`, `yarray`, `xtype`


bobleesj:
### Problem

Continuing our discussion from https://github.com/diffpy/diffpy.utils/issues/224 - We do not want User to have an empty `DiffractionObject` instance. 

### Proposed solution

In our `init` method, we could modfy

from:

```python
class DiffractionObject:
    def input_data(
        self,
        xarray,
        yarray,
        xtype,
        metadata={},
        scat_quantity=None,
        name=None,
        wavelength=None,
    ):

```

to: 

```python
class DiffractionObject:
    def __init__(
        self,
        xarray,
        yarray,
        xtype,
        wavelength,
        scat_quantity="",
        name="",
        metadata={},
    ):

        self._id = uuid.uuid4()
        self.input_data(xarray, yarray, xtype, wavelength, scat_quantity, name, metadata)
```

Here, User needs to provide `xarray`, `yarray`, `xtype` and `wavelength`. Here, `xtype` and `wavelneght` are required for  the ` self._set_xarrays(xarray, xtype)` method within `input_data ` where we utilize `self.wavelength` and `self.xtype` as shown below:

```python
    def _set_xarrays(self, xarray, xtype):
        self._all_arrays = np.empty(shape=(len(xarray), 4))
        if xtype.lower() in QQUANTITIES:
            self._all_arrays[:, 1] = xarray
            self._all_arrays[:, 2] = q_to_tth(xarray, self.wavelength)
            self._all_arrays[:, 3] = q_to_d(xarray)
    ....
```

@sbillinge I have a draft PR here: https://github.com/diffpy/diffpy.utils/pull/228 for this implementation.

sbillinge:
I think that's about right, and I think we want `input_arrays` to be private. 

We don't want wavelength required though. Set default to none. I think we handle that in thighs arrays. Also, pretty soon Caden will need tof neutron data in the DO and then angle is fixed and lamba is in the array....

Actually, kind of interestingly, because he has lots of detectors, angle and wavelength could both be arrays.  Let's leave that for a nice future project for Caden, but for now make wavelength optional.

PR #228, opened 2024-12-14
Do not allow an empty instance of DiffractionObject - require `xarrays` `yarrays` `xtype`

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/227


src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,68 @@ def _setter_wmsg(attribute):
...
-        if metadata is None:
-            metadata = {}
-        self.metadata = metadata
-        if xtype is None:
-            xtype = ""
+
+        self._id = uuid.uuid4()
+        self.input_data(xarray, yarray, xtype, wavelength, scat_quantity, name, metadata)
+
+    def input_data(self, xarray, yarray, xtype, wavelength, scat_quantity="", name="", metadata={}):
```
bobleesj, line 52: having `input_data` method as the first method right below `__init__`

src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,68 @@ def _setter_wmsg(attribute):
...
+        """
+        Insert a new scattering quantity into the scattering object.
```
bobleesj, line 54: Improved docstrings, previous version:

```
        f"""
        insert a new scattering quantity into the scattering object
        Parameters
        ----------
        xarray array-like of floats
          the independent variable array
        yarray array-like of floats
          the dependent variable array
        xtype string
          the type of quantity for the independent variable from {*XQUANTITIES, }
        metadata, scat_quantity, name and wavelength are optional.  They have the same
        meaning as in the constructor. Values will only be overwritten if non-empty values are passed.
        Returns
        -------
        Nothing.  Updates the object in place.
        """
```

tests/conftest.py
```diff
@@ -28,3 +31,9 @@ def _load(filename):
         return base_path / filename
 
     return _load
+
+
+@pytest.fixture
+def do_minimal():
```
bobleesj, line 37: Setup a reusable `do` object under `conftest.py` Calling it `do_minimal` since it has the minimal number of arguments and setup of `xarray`, `yarray`, `xtype` and `wavelength`

src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,67 @@ def _setter_wmsg(attribute):
 
 class DiffractionObject:
     def __init__(
-        self, name=None, wavelength=None, scat_quantity=None, metadata=None, xarray=None, yarray=None, xtype=None
+        self,
+        xarray,
+        yarray,
+        xtype,
+        wavelength,
+        scat_quantity="",
```
sbillinge, line 44: See discussion in issue about wavelength.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,67 @@ def _setter_wmsg(attribute):
...
-            name = ""
-        self.name = name
-        if metadata is None:
-            metadata = {}
-        self.metadata = metadata
-        if xtype is None:
-            xtype = ""
+
+        self._id = uuid.uuid4()
+        self.input_data(xarray, yarray, xtype, wavelength, scat_quantity, name, metadata)
```
sbillinge, line 50: Make private?

src/diffpy/utils/diffraction_objects.py
sbillinge: If we make it private we may want to move the docstring to the constructor?  Also, then change "insert a new scattering object blah blah, to instantiate a new scattering object or sthg.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,67 @@ def _setter_wmsg(attribute):
...
+
+    def input_data(self, xarray, yarray, xtype, wavelength, scat_quantity="", name="", metadata={}):
+        """
+        Insert a new scattering quantity into the scattering object.
+
+        Parameters
+        ----------
+        xarray : array-like
+            The independent variable array (e.g., "q", "tth", or "d").
+        yarray : array-like
```
sbillinge, line 60: E.g, containing ...., or "d" values?

src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,67 @@ def _setter_wmsg(attribute):
...
+            The dependent variable array corresponding to intensity values.
+        xtype : str
+            The type of the independent variable in `xarray`. Must be one of {*XQUANTITIES},
+            such as "q", "tth", or "d".
```
sbillinge, line 64: Remove this line. It is handled on the line above.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,67 @@ def _setter_wmsg(attribute):
...
+        wavelength : float
+            The wavelength of the incoming beam, specified in angstroms (Å).
+        scat_quantity : str, optional
+            The type of scattering experiment (e.g., "x-ray", "neutron"). Default is an empty string "".
+        name : str, optional
+            The name or label for the scattering data. Default is an empty string "".
+        metadata : dict, optional
+            The additional metadata associated with the diffraction object. Default is {}.
```
sbillinge, line 72: I would give an example also of a non empty one to help non python people to get the syntax right

src/diffpy/utils/diffraction_objects.py
```diff
@@ -36,26 +36,67 @@ def _setter_wmsg(attribute):
...
-        if xarray is None:
-            xarray = np.empty(0)
-        if yarray is None:
-            yarray = np.empty(0)
-
-        self._id = uuid.uuid4()
-        self.input_data(xarray, yarray, xtype)
+        self._input_xtype = xtype
+        self._set_xarrays(xarray, xtype)
+        self._all_arrays[:, 0] = yarray
```
sbillinge, line 99: It would be cleaner of this was moved into _set_arrays unless there was a reason not to do that because it is reused somewhere or something

tests/conftest.py
sbillinge: I like this name. I suggested test_do or something, but your name is better

tests/test_diffraction_objects.py
```diff
@@ -41,17 +36,17 @@
         {
             "name": "something",
             "scat_quantity": "",
-            "wavelength": None,
-            "xtype": "",
+            "wavelength": 0.71,
```
sbillinge, line 39: Return wavelengths back to None or absent to test that behavior.

tests/test_diffraction_objects.py
```diff
@@ -241,40 +236,6 @@ def test_dump(tmp_path, mocker):
 
 
 tc_params = [
-    (
-        {},
```
sbillinge, line 245: Where are these things now getting tested?

sbillinge (commented):
Please see inline.  We may want to check it is ok for us to have things like empty string details instead of None that converts to empty string

bobleesj:
> Please see inline. We may want to check it is ok for us to have things like empty string details instead of None that converts to empty string

@sbillinge Before making PRs, when would be beneficial to have  `none` for `xtype` or `scat_quantity`?

My understanding is that we usually want to pass typed variables so that typecasting can be minimal.

sbillinge:
> > Please see inline. We may want to check it is ok for us to have things like empty string details instead of None that converts to empty string
> 
> @sbillinge Before making PRs, when would be beneficial to have `none` for `xtype` or `scat_quantity`?
>
I can't think of  a case where we want none for xtype.  I think we require this, right?  For scat_quantity, the tradeoff is between making these easier to use (and therefore encouraging adoption) vs. making them stricter so we collect better metadata.  I think leaning in the direction of adoption may be a good starting point for that (though if they are adopted, we may regret that later).  That is the UC for a None default for scat_quantity: lazy users....

> 
> My understanding is that we usually want to pass typed variables so that typecasting can be minimal.
> 
The issue is with setting mutable objects as defaults.  In that case you should pass None and then you have logic like, `if thing is None: thing = []` or whatever.   I was being a bit too conservative about the empty strings.  i was thinking they are mutable since they can act like lists, but I checked and they are not subject to this issue.  So we could directly set empty strings (and of course floats and ints) as defaults.

> Extra comment:
> 
> I can see that having `wavelength=none` makes sense since we don't want a float value associated when it's not provided.
>
yes, in this case None is different from 0.  We want None here.

src/diffpy/utils/diffraction_objects.py
bobleesj: done

src/diffpy/utils/diffraction_objects.py
bobleesj: done

src/diffpy/utils/diffraction_objects.py
bobleesj: done

src/diffpy/utils/diffraction_objects.py
bobleesj: done

src/diffpy/utils/diffraction_objects.py
bobleesj: done

src/diffpy/utils/diffraction_objects.py
bobleesj: done

src/diffpy/utils/diffraction_objects.py
bobleesj: done

tests/conftest.py
bobleesj: done

src/diffpy/utils/diffraction_objects.py
```diff
@@ -35,27 +35,82 @@ def _setter_wmsg(attribute):
...
+                "Please re-initialize 'DiffractionObject' or re-run the method 'input_data' "
+                "with 'xarray' and 'yarray' of identical length."
+            )
+
+        self.scat_quantity = scat_quantity
+        self.wavelength = wavelength
+        self.metadata = metadata
+        self.name = name
+        self._input_xtype = xtype
+        self._set_arrays(xarray, xtype, yarray)
```
bobleesj, line 112: as discussed - private func

src/diffpy/utils/diffraction_objects.py
```diff
@@ -35,27 +35,82 @@ def _setter_wmsg(attribute):
...
+    >>> do = DiffractionObject(
+    ...     xarray=x,
+    ...     yarray=y,
+    ...     xtype="q",
+    ...     wavelength=1.54,
+    ...     scat_quantity="x-ray",
+    ...     metadata=metadata
+    ... )
+    >>> print(do.metadata)
+    """
```
bobleesj, line 79: Screenshot 2024-12-14 at 1 44 47 PM

It creates a copy-pastable docstring for new users.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -320,70 +375,15 @@ def _set_xarrays(self, xarray, xtype):
             self._all_arrays[:, 3] = xarray
             self._all_arrays[:, 1] = d_to_q(xarray)
             self._all_arrays[:, 2] = d_to_tth(xarray, self.wavelength)
+
+    def _set_min_max_xarray(self):
```
bobleesj, line 379: `_set_arrays` is used to set `all_arrays` and I thought it would be more modular to have `_set_min_max_xarray` for determining min/max values of q, tth, d.

tests/test_diffraction_objects.py
```diff
@@ -99,16 +84,16 @@
     (  # Different scat_quantity
         {
             "scat_quantity": "x-ray",
-            "wavelength": None,
```
bobleesj, line 102: since we are testing the user behavior, it is most likely that User wouldn't enter `none` right?

src/diffpy/utils/diffraction_objects.py
bobleesj: (This PR is getting bigger..)

tests/test_diffraction_objects.py
```diff
@@ -11,16 +11,10 @@
 from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
 
 params = [
-    (  # Default
```
bobleesj, line 14: These invalid init test cases are provided below.

tests/test_diffraction_objects.py
bobleesj: pls see `test_init_invalid_args` below

tests/test_diffraction_objects.py
```diff
@@ -487,13 +427,39 @@ def test_dump(tmp_path, mocker):
...
+)
+def test_init_valid(init_args, expected_do_dict):
+    actual_do_dict = DiffractionObject(**init_args).__dict__
+    diff = DeepDiff(
+        actual_do_dict, expected_do_dict, ignore_order=True, significant_digits=13, exclude_paths="root['_id']"
+    )
     assert diff == {}
 
 
+test_init_invalid_params = [
```
bobleesj, line 442: `xarray` `yarrays` and `xtype` requires - test cases when these are not provided during `init`

bobleesj (commented):
@sbillinge This is ready for review  - pls see in-line comments..

bobleesj:
(Just added news)

bobleesj (commented):
@sbillinge re-revisiting after resolving a `conftest.py` conflict. and ready for review.

sbillinge (commented):
I reviewed this yesterday or so but it looks as if I forgot to submit the review.....sorry.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -35,27 +35,82 @@ def _setter_wmsg(attribute):
...
+    --------
+    Create a DiffractionObject for X-ray scattering data:
+
+    >>> import numpy as np
+    >>> from diffpy.utils.diffraction_objects import DiffractionObject
+    ...
+    >>> x = np.array([0.12, 0.24, 0.31, 0.4])  # independent variable (e.g., q)
+    >>> y = np.array([10, 20, 40, 60])  # intensity values
+    >>> metadata = {
+    ...     "package_info": {"version": "3.6.0"}
```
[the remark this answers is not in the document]
bobleesj: Screenshot 2024-12-15 at 9 12 12 PM

Done

src/diffpy/utils/diffraction_objects.py
```diff
@@ -35,27 +35,82 @@ def _setter_wmsg(attribute):
...
+            raise ValueError(_xtype_wmsg(xtype))
+
+        # Check xarray and yarray have the same length
+        if len(xarray) != len(yarray):
+            raise ValueError(
+                "'xarray' and 'yarray' must have the same length. "
+                "Please re-initialize 'DiffractionObject' or re-run the method 'input_data' "
+                "with 'xarray' and 'yarray' of identical length."
+            )
+
```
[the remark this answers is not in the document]
bobleesj: done

src/diffpy/utils/diffraction_objects.py
```diff
@@ -35,27 +35,82 @@ def _setter_wmsg(attribute):
...
+    >>> metadata = {
+    ...     "package_info": {"version": "3.6.0"}
+    ... }
+    >>> do = DiffractionObject(
+    ...     xarray=x,
+    ...     yarray=y,
+    ...     xtype="q",
+    ...     wavelength=1.54,
+    ...     scat_quantity="x-ray",
+    ...     metadata=metadata
```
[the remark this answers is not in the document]
bobleesj: done

src/diffpy/utils/diffraction_objects.py
```diff
@@ -35,27 +35,82 @@ def _setter_wmsg(attribute):
...
-            yarray = np.empty(0)
 
         self._id = uuid.uuid4()
-        self.input_data(xarray, yarray, xtype)
+        self._input_data(xarray, yarray, xtype, wavelength, scat_quantity, name, metadata)
+
+    def _input_data(self, xarray, yarray, xtype, wavelength, scat_quantity, name, metadata):
+        if xtype not in XQUANTITIES:
+            raise ValueError(_xtype_wmsg(xtype))
+
```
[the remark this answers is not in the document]
bobleesj: done

src/diffpy/utils/diffraction_objects.py
```diff
@@ -298,16 +353,16 @@ def get_array_index(self, value, xtype=None):
...
-            xtype = self._input_xtype
+        xtype = self._input_xtype
         array = self.on_xtype(xtype)[0]
         if len(array) == 0:
             raise ValueError(f"The '{xtype}' array is empty. Please ensure it is initialized.")
         i = (np.abs(array - value)).argmin()
         return i
 
-    def _set_xarrays(self, xarray, xtype):
+    def _set_arrays(self, xarray, xtype, yarray):
```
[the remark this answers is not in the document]
bobleesj: done

tests/test_diffraction_objects.py
bobleesj: @sbillinge I think these aren't valid since we `DiffractionObject` requires x, y, and xtype (current PR)

src/diffpy/utils/diffraction_objects.py
```diff
@@ -231,16 +288,16 @@ def get_array_index(self, value, xtype=None):
...
-            xtype = self._input_xtype
+        xtype = self._input_xtype
         array = self.on_xtype(xtype)[0]
         if len(array) == 0:
             raise ValueError(f"The '{xtype}' array is empty. Please ensure it is initialized.")
         i = (np.abs(array - value)).argmin()
         return i
 
-    def _set_xarrays(self, xarray, xtype):
+    def _set_arrays(self, xarray, yarray, xtype):
```
bobleesj, line 298: as suggested - ordered by x, y, xtype

bobleesj:
@sbillinge ready for review!

sbillinge (approved):

sbillinge:
There is a lot of good stuff on here and this is really getting wrestled into shape in my mind so I want to merge this (and will).  There were two things left to do that I will mae issues for.  Thanks for your work on this.  It is really great!

Issue #229, opened 2024-12-14
Refactor `test_on_xtype()` using `@pytest.mark.parametrize` and `conftest.py`


bobleesj:
### Problem

We have 2 PRs opened at the moment, working on the same file.

Once https://github.com/diffpy/diffpy.utils/pull/228 is addressed and merged, refactor:

```python
def test_on_xtype():
    do = DiffractionObject(wavelength=2 * np.pi, xarray=np.array([30, 60]), yarray=np.array([1, 2]), xtype="tth")
    assert np.allclose(do.on_xtype("tth"), [np.array([30, 60]), np.array([1, 2])])
    assert np.allclose(do.on_xtype("2theta"), [np.array([30, 60]), np.array([1, 2])])
    assert np.allclose(do.on_xtype("q"), [np.array([0.51764, 1]), np.array([1, 2])])
    assert np.allclose(do.on_xtype("d"), [np.array([12.13818, 6.28319]), np.array([1, 2])])
```

Ref: Continuing our in-line comment interactions from https://github.com/diffpy/diffpy.utils/pull/225#discussion_r1884798351

### Proposed solutions

- Use well defined reusable DO instances defined under `contest.py`

PR #231, opened 2024-12-14
Refactor `test-on-xtype` test function 

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/229 - Refactor test_on_xtype() using @pytest.mark.parametrize and conftest.py

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,10 @@
 import json
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


bobleesj:
@sbillinge Ready for review - probably better to review this https://github.com/diffpy/diffpy.utils/pull/228 first..

PR #233, opened 2024-12-14
Remove `set_angles_from_list`, `set_angles_from`, `set_qs_from_range` 

### LINKED ISSUE #232: Remove `set_angles_from_list`, `set_angles_from`, `set_qs_from_range`  and tests associated
@sbillinge Can I remove these util functions from the package? Since @yucongalicechen has started writing docs and docstrings, thought it would be best to keep it as portable as possible. But, I have copied and pasted those test functions explicitly above.

_Originally posted by @bobleesj in https://github.com/diffpy/diffpy.utils/issues/198#issuecomment-2543322377_

---

Remove the functions provided in the issue attached.

bobleesj:
Closes #232


src/diffpy/utils/diffraction_objects.py
```diff
@@ -215,73 +215,6 @@ def id(self):
     def id(self, _):
         raise AttributeError(_setter_wmsg("id"))
 
-    def set_angles_from_list(self, angles_list):
```
bobleesj, line 218: For the following methods, I've double checked that no tests associated were found and pytest runs locally without these.

bobleesj (commented):
@sbillinge ready for review

PR #250, opened 2024-12-18
implement a skip_config_creation option in get_user_info

### LINKED ISSUE #244: implement a `skip_config_creation` option in `get_user_info`
### Problem

### Proposed solution

```

an input file, type

.. code-block:: python

```

yucongalicechen:
closes #244 
@sbillinge ready for some feedback
btw I see in the docs you used `get_user_data` instead of `get_user_info` at a couple places, which seems to be a better name, shall we change to `get_user_data` instead?

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -22,6 +22,9 @@ def _run_tests(inputs, expected):
     config = get_user_info(args)
     assert config.get("username") == expected_username
     assert config.get("email") == expected_email
+    config = get_user_info(args, skip_config_creation=True)
+    assert config.get("username") == expected_username
+    assert config.get("email") == expected_email
 
 
 params_user_info_with_home_conf_file = [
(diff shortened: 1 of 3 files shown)


tests/test_tools.py
```diff
@@ -22,6 +22,9 @@ def _run_tests(inputs, expected):
     config = get_user_info(args)
     assert config.get("username") == expected_username
     assert config.get("email") == expected_email
+    config = get_user_info(args, skip_config_creation=True)
+    assert config.get("username") == expected_username
+    assert config.get("email") == expected_email
```
yucongalicechen, line 27: It looks like we can reuse this run_test function for most test cases for different `skip_config_creation` except when there're no inputs or files, so I added this here

sbillinge:
> closes #244 @sbillinge ready for some feedback btw I see in the docs you used `get_user_data` instead of `get_user_info` at a couple places, which seems to be a better name, shall we change to `get_user_data` instead?

haha, that was a mistake on my part.    I am not sure tbh.  I am ok with either "user-data* or "user-info".   @bobleesj @yucongalicechen which do you think is more obvious?

tests/test_tools.py
```diff
@@ -118,7 +121,20 @@ def test_get_user_info_no_conf_file_no_inputs(monkeypatch, inputsa, inputsb, exp
...
+    args = {"username": inputsa[0], "email": inputsa[1]}
+    expected_username, expected_email = expected
+
+    # Test with user inputs
+    config = get_user_info(args)
+    assert config.get("username") == expected_username
+    assert config.get("email") == expected_email
+
+    # Test skipping config creation, expecting None values
+    config = get_user_info(args, skip_config_creation=True)
```
sbillinge, line 134: I am a little confused here....isn't this getting `args` as written?

sbillinge (commented):
tbh I find the tests confusing for the ``get_user_info/data`` workflow.  Can we maybe have them written following the new standards that @bobleesj and I worked out?

yucongalicechen:
closed as replaced by #253

PR #253, opened 2024-12-19
change args to user_info and fix docstring

### LINKED ISSUE #244: implement a `skip_config_creation` option in `get_user_info`

### LINKED ISSUE #245: I think the `get_user_info` docs don't match the code
### Problem
The docs don't seem to correspond to the code.

I was tweaking the docs and checking the code.  Places where I didn't update the docs it is because I think the docs are better behavior than the code....

### Proposed solution
update the code to match the docs (with tests).

One thing I remember concretely is that `get_user_info` gets an optional argument of `args` that is type argparse.Namespace but the docs say a dict can be passed.

Since we decided to make this a user-facing functionality that can be used by others, I think it is more reusable if it takes a dict and not the full args Namespace?

But there may be other things, so have someone go through the docs and type all the commands illustrated in there and make sure the code actually does what it says.

yucongalicechen:
closes #245, closes #244
@sbillinge ready for review

### DIFF
--- a/src/diffpy/utils/tools.py
+++ b/src/diffpy/utils/tools.py
@@ -85,7 +85,16 @@ def _sorted_merge(*dicts):
     return merged
 
 
+def _get_config_info(user_info):
+    global_config = _load_config(Path().home() / "diffpyconfig.json")
+    local_config = _load_config(Path().cwd() / "diffpyconfig.json")
+    if global_config or local_config:
+        return _sorted_merge(global_config, local_config, user_info)
+    return None
+
+
 def _create_global_config(user_info):
+    print(user_info_imsg)
     username = input(
         f"Please enter the name you would want future work to be credited to "
         f"[{user_info.get('username', '')}]:  "
@@ -108,14 +117,13 @@ def _create_global_config(user_info):
 
 def get_user_info(user_info=None, skip_config_creation=False):
     """
-    Get username and email configuration.
+    Retrieve or create user configuration (username and email).
 
     Workflow:
-    1. First attempts to load config file from global and local paths.
-    2. If config files exist, prioritizes values from user_info, then local config, then global config.
-    3. Otherwise, if user wants to skip config creation, uses user_info as the final config, even if it's empty.
-    4. Otherwise, if no config files exist and user does not want to skip config creation, prompts for user inputs.
-    5. If no config files exist, a global config file will only be created if both username and email are valid.
+    1. Loads config from global and local paths, prioritizing user_info, then local, then global.
+    2. If no config files are found and skip_config_creation is False, prompts the user for input.
+       Creates a global config file only if both username and email are valid.
+    3. If no config files and skip_config_creation is True, uses user_info (even if empty).
 
     Parameters
     ----------
@@ -127,17 +135,15 @@ def get_user_info(user_info=None, skip_config_creation=False):
     dict or None:
         The dictionary containing username and email with corresponding values.
     """
-    global_config = _load_config(Path().home() / "diffpyconfig.json")
-    local_config = _load_config(Path().cwd() / "diffpyconfig.json")
-    if global_config or local_config:
-        return _sorted_merge(global_config, local_config, user_info)
-    if skip_config_creation:
-        return {
-            "username": _stringify(user_info.get("username", "")),
-            "email": _stringify(user_info.get("email", "")),
-        }
-    print(user_info_imsg)
-    return _create_global_config(user_info)
+    config = _get_config_info(user_info)
+    if config:
+        return config
+    if not skip_config_creation:
+        return _create_global_config(user_info)
+    return {
+        "username": _stringify(user_info.get("username", "")),
+        "email": _stringify(user_info.get("email", "")),
+    }
 
 
 def get_package_info(package_names, metadata=None):
--- a/src/diffpy/utils/tools.py
+++ b/src/diffpy/utils/tools.py
@@ -10,7 +10,7 @@ user_info_imsg = (
     "computer and your name will be automatically associated with subsequent diffpy data by default.\n"
     "This is not recommended on a shared or public computer. "
     "You will only have to do that once.\n"
-    "For more information, please refer to www.diffpy.org/diffpy.utils/examples/toolsexample.html"
+    "For more information, please refer to www.diffpy.org/diffpy.utils/examples/toolsexample.html."
 )
 
 
@@ -36,7 +36,7 @@ def _clean_dict(obj):
     return obj
 
 
-def stringify(obj):
+def _stringify(obj):
     """
     Convert None to an empty string.
 
@@ -53,7 +53,7 @@ def stringify(obj):
     return obj if obj is not None else ""
 
 
-def load_config(file_path):
+def _load_config(file_path):
     """
     Load configuration from a .json file.
 
@@ -77,7 +77,7 @@ def load_config(file_path):
         return None
 
 
-def _merge_sorted_configs(*dicts):
+def _sorted_merge(*dicts):
     merged = {}
     for d in dicts:
         d = _clean_dict(d)
@@ -93,15 +93,16 @@ def _create_global_config(user_info):
     email = input(f"Please enter the your email " f"[{user_info.get('email', '')}]:  ").strip() or user_info.get(
         "email", ""
     )
-    config = {"username": stringify(username), "email": stringify(email)}
+    config = {"username": _stringify(username), "email": _stringify(email)}
     if username and email:
         with open(Path().home() / "diffpyconfig.json", "w") as f:
             f.write(json.dumps(config))
-    print(
-        f"You can manually edit the config file at {Path().home() / 'diffpyconfig.json'} using any text editor.\n"
-        f"Or you can update the config file by passing new values to get_user_info(), "
-        f"see examples here: https://www.diffpy.org/diffpy.utils/examples/toolsexample.html"
-    )
+        print(
+            f"You can manually edit the config file at "
+            f"{Path().home() / 'diffpyconfig.json'} using any text editor.\n"
+            f"Or you can update the config file by passing new values to get_user_info(), "
+            f"see examples here: https://www.diffpy.org/diffpy.utils/examples/toolsexample.html."
+        )
     return config
 
 
@@ -109,12 +110,12 @@ def get_user_info(user_info=None, skip_config_creation=False):
     """
     Get username and email configuration.
 
-    The workflow is following:
-    We first attempt to load config file from global and local paths.
-    If any exists, it prioritizes values from user_info, then local, then global.
-    Otherwise, if user wants to skip config creation, it uses user_info as the final info, even if it's empty.
-    Otherwise, prompt for user inputs, and create a global config file.
-    Removes invalid global config file if creation is needed, replacing it with empty username and email.
+    Workflow:
+    1. First attempts to load config file from global and local paths.
+    2. If config files exist, prioritizes values from user_info, then local config, then global config.
+    3. Otherwise, if user wants to skip config creation, uses user_info as the final config, even if it's empty.
+    4. Otherwise, if no config files exist and user does not want to skip config creation, prompts for user inputs.
+    5. If no config files exist, a global config file will only be created if both username and email are valid.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/tools.py
```diff
@@ -75,11 +75,14 @@ def _sorted_merge(*dicts):
...
     username = input(
-        f"Please enter the name you would want future work to be credited to " f"[{args.get('username', '')}]:  "
-    ).strip() or args.get("username", "")
-    email = input(f"Please enter the your email " f"[{args.get('email', '')}]:  ").strip() or args.get("email", "")
+        f"Please enter the name you would want future work to be credited to "
+        f"[{user_info.get('username', '')}]:  "
+    ).strip() or user_info.get("username", "")
+    email = input(f"Please enter the your email " f"[{user_info.get('email', '')}]:  ").strip() or user_info.get(
+        "email", ""
+    )
```
yucongalicechen, line 85: We're already writing and testing everything in dict not args I think

sbillinge:
I am still struggling a bit with this functionality.  I think maybe I got to the botto of why this is the case (though maybe not!).  I think the function is trying to do two things which is making me confused.  The two things are (a) getting the config information from config files and (b) a workflow to create/maintain the config file.  I wonder if this all gets a bit easier if we separate these behaviorws, for example, with a conditional, but this taking place outside the `get_info` function?   Maybe we could sketch out this flow in a diagram?

yucongalicechen:
> I am still struggling a bit with this functionality. I think maybe I got to the botto of why this is the case (though maybe not!). I think the function is trying to do two things which is making me confused. The two things are (a) getting the config information from config files and (b) a workflow to create/maintain the config file. I wonder if this all gets a bit easier if we separate these behaviorws, for example, with a conditional, but this taking place outside the `get_info` function? Maybe we could sketch out this flow in a diagram?

Yes, I agree we can make it a bit more readable! I drew a diagram on this. I am thinking that we can have one private function for each action in the diagram?

workflow

yucongalicechen:
@sbillinge Since we're redoing the workflow anyways, I'm adding the skip_config_creation for #244 here too. Here's a new diagram I'm thinking about:

workflow2

yucongalicechen:
^push an intermediate commit in case anyone would like to comment, I will likely reformat some functions and make the comments clearer

yucongalicechen:
I edited the codes again and it's ready for some feedback now @sbillinge

src/diffpy/utils/tools.py
```diff
@@ -1,11 +1,20 @@
...
+    "By following the prompts below you can add your name and email to this file on the current "
+    "computer and your name will be automatically associated with subsequent diffpy data by default.\n"
+    "This is not recommended on a shared or public computer. "
+    "You will only have to do that once.\n"
+    "For more information, please refer to www.diffpy.org/diffpy.utils/examples/toolsexample.html."
+)
 
-def clean_dict(obj):
+
+def _clean_dict(obj):
```
yucongalicechen, line 17: make this function private

src/diffpy/utils/tools.py
```diff
@@ -71,67 +80,64 @@ def load_config(file_path):
 def _sorted_merge(*dicts):
     merged = {}
     for d in dicts:
+        d = _clean_dict(d)
```
yucongalicechen, line 83: add a step here so we don't need to pass `_clean_dict(d)` for every `d`

src/diffpy/utils/tools.py
```diff
@@ -71,67 +80,64 @@ def load_config(file_path):
...
-        f.write(json.dumps({"username": stringify(username), "email": stringify(email)}))
-    print(
-        f"You can manually edit the config file at {Path().home() / 'diffpyconfig.json'} using any text editor.\n"
-        f"Or you can update the config file by passing new values to get_user_info(), "
-        f"see examples here: https://www.diffpy.org/diffpy.utils/examples/toolsexample.html"
+        f"Please enter the name you would want future work to be credited to "
+        f"[{user_info.get('username', '')}]:  "
+    ).strip() or user_info.get("username", "")
+    email = input(f"Please enter the your email " f"[{user_info.get('email', '')}]:  ").strip() or user_info.get(
+        "email", ""
```
yucongalicechen, line 94: name change from `args` to `user_info`

src/diffpy/utils/tools.py
```diff
@@ -71,67 +80,64 @@ def load_config(file_path):
...
     )
-    return return_bool
+    config = {"username": _stringify(username), "email": _stringify(email)}
+    if username and email:
```
yucongalicechen, line 97: The function originally returns a bool that passes back to `get_user_info` to see if we need to create a global config file. If user only input none or one of username / email, an empty global config file will be created. I simplified the logic here so that we only create a global config file if both username and email are present. Does this make sense though?

tests/test_tools.py
```diff
@@ -16,109 +16,201 @@ def _setup_dirs(monkeypatch, user_filesystem):
     return home_dir
 
 
-def _run_tests(inputs, expected):
-    args = {"username": inputs[0], "email": inputs[1]}
-    expected_username, expected_email = expected
-    config = get_user_info(args)
+def _run_tests(cli_inputs, expected):
+    user_info = {"username": cli_inputs["cli_username"], "email": cli_inputs["cli_email"]}
```
yucongalicechen, line 20: "cli_inputs" refer to function arguments passed to `get_user_info`. Not sure if this is the best name here but I want to distinguish them from user inputs via prompts ("inputs").

tests/test_tools.py
```diff
@@ -16,109 +16,201 @@ def _setup_dirs(monkeypatch, user_filesystem):
...
     _setup_dirs(monkeypatch, user_filesystem)
     os.remove(Path().home() / "diffpyconfig.json")
-    inp_iter = iter(inputsb)
+    inp_iter = iter([inputs["input_username"], inputs["input_email"]])
     monkeypatch.setattr("builtins.input", lambda _: next(inp_iter))
-    _run_tests(inputsa, expected)
+    _run_tests(cli_inputs, expected)
     confile = Path().home() / "diffpyconfig.json"
-    assert confile.is_file()
+    assert confile.is_file() == expected["config_file_exists"]
```
yucongalicechen, line 206: global config file should only be created when both username and email are given

tests/test_tools.py
```diff
@@ -16,109 +16,201 @@ def _setup_dirs(monkeypatch, user_filesystem):
...
+@pytest.mark.parametrize("cli_inputs, expected", params_user_info_no_conf_file_no_inputs)
+def test_get_user_info_no_conf_file_no_inputs(monkeypatch, cli_inputs, expected, user_filesystem):
     _setup_dirs(monkeypatch, user_filesystem)
     os.remove(Path().home() / "diffpyconfig.json")
-    inp_iter = iter(inputsb)
-    monkeypatch.setattr("builtins.input", lambda _: next(inp_iter))
-    _run_tests(inputsa, expected)
+    _run_tests(cli_inputs, expected)
     confile = Path().home() / "diffpyconfig.json"
     assert confile.exists() is False
```
yucongalicechen, line 215: config file is never created if user only provides function arguments

tests/test_tools.py
bobleesj: Do we want to still have a helper function called `_run_tests` within each test func? 

The name `run_tests` to me is quite ambiguous - what test is it running? comparing username and email?

tests/test_tools.py
```diff
@@ -16,109 +16,201 @@ def _setup_dirs(monkeypatch, user_filesystem):
...
+def _run_tests(cli_inputs, expected):
+    user_info = {"username": cli_inputs["cli_username"], "email": cli_inputs["cli_email"]}
+    config = get_user_info(user_info=user_info, skip_config_creation=cli_inputs["skip_config_creation"])
+    expected_username = expected["expected_username"]
+    expected_email = expected["expected_email"]
     assert config.get("username") == expected_username
     assert config.get("email") == expected_email
 
 
 params_user_info_with_home_conf_file = [
```
bobleesj, line 28: `params_user_info_with_home_conf_file` may not be needed - please see our discussion: https://github.com/diffpy/diffpy.utils/pull/236/files#r1885496247 @yucongalicechen

tests/test_tools.py
```diff
@@ -16,109 +16,201 @@ def _setup_dirs(monkeypatch, user_filesystem):
...
     confile = Path().home() / "diffpyconfig.json"
-    assert confile.is_file()
+    assert confile.is_file() == expected["config_file_exists"]
 
 
-@pytest.mark.parametrize("inputsa, inputsb, expected", params_user_info_no_conf_file_no_inputs)
-def test_get_user_info_no_conf_file_no_inputs(monkeypatch, inputsa, inputsb, expected, user_filesystem):
+@pytest.mark.parametrize("cli_inputs, expected", params_user_info_no_conf_file_no_inputs)
+def test_get_user_info_no_conf_file_no_inputs(monkeypatch, cli_inputs, expected, user_filesystem):
     _setup_dirs(monkeypatch, user_filesystem)
```
bobleesj, line 211: If you could help me understanding - what's the role of `moneypatch` and why do we need to setup a dir using `_setup_dirs`?

sbillinge:
I thought about this more and, as I mentioned before, I think it makes more sense to disentangle the two workflows (1) get_user_info and (2) update_config.  I sketched this in two images that I paste below.  I don't necessarily expect you to be able to read them, but they show the structure.

The starting point of the UC is labpdfproc (LPP), for example, first running an `update_global_config` workflow, followed by a separate `build_metadata` workflow, i.e.,
PXL_20241220_105549384 MP

where the `build_metadata` workflow looks like
PXL_20241220_105541819

The `build_metadata` workflow is as we have now, runtime > local config > global config for uname, email, orcid.

The `update_global_config` checks for `global_config`.  If the file is missing or if unam/email/orcid are missing in the file, it runs something like  `if not skip_config` it prompts users for input and creates/updates the config file.

I think separating these makes them more reusable.  I want to start using this (and the get_package_info) in all our codes so we start the arduous process of collecting better metadata.

tests/test_tools.py
```diff
@@ -16,109 +16,201 @@ def _setup_dirs(monkeypatch, user_filesystem):
...
+    (
+        {"cli_username": None, "cli_email": None, "skip_config_creation": False},
+        {"expected_username": "home_username", "expected_email": "home@email.com"},
+    ),
+    (
+        {"cli_username": "cli_username", "cli_email": None, "skip_config_creation": False},
+        {"expected_username": "cli_username", "expected_email": "home@email.com"},
+    ),
+    (
+        {"cli_username": None, "cli_email": "cli@email.com", "skip_config_creation": False},
```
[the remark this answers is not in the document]
yucongalicechen: Hey Bob, thanks for the feedback!! I think these repeated key values correspond to different test cases. Basically this function tries to get username/email from four places: (a) prompt input, (b) function arguments (which we called cli input because it was initially used by labpdfproc, where the function arguments are from cli), (c) local config file, (d) global config file, with priority order.

Currently we have four test functions that correspond to:
(1) we only have global config file, function prioritizes (b) to (d) (that is, replace (d) if (b) is non-empty).
(2) we have local config file with/without global config file, function prioritizes (b) to (c).
(3) we dont have config files, and user does not choose to skip config creation, then function prioritizes (a) to (b), and create a global config file if both username/email arguments are present.
(4) we dont have config files, and user chooses to skip config creation (this means we don't have (a)), then the function reads (b), and replace username/email with an empty string if no value given.

So I think all arguments and test values are necessary here. I will add a bit more comment to the tests so that it's clearer. We might be able to find a better way to organize the tests?

tests/test_tools.py
yucongalicechen: I think this is probably a different test case than that issue..?

tests/test_tools.py
yucongalicechen: this is to set up the global and local file paths for the test cases because the function will try to read files from these filepaths

tests/test_tools.py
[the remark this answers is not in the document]
bobleesj: Thanks @yucongalicechen for the comments.  Yeah I had a quick look at the src code. 

Username/email from 4 places:

1. prompt
2. function args
3. local config
4. global config

and the test functions are basically testing the input/out and reading/creating config files as needed.. 

Noted.

src/diffpy/utils/tools.py
```diff
@@ -71,67 +80,70 @@ def load_config(file_path):
...
-    return config
+    config = _get_config_info(user_info)
+    if config:
+        return config
+    if not skip_config_creation:
+        return _create_global_config(user_info)
+    return {
+        "username": _stringify(user_info.get("username", "")),
+        "email": _stringify(user_info.get("email", "")),
+    }
```
yucongalicechen, line 146: simplified the workflow here

tests/test_tools.py
yucongalicechen: Whoops missed this comment. Yeah I think we can keep this test  but maybe have a better name

yucongalicechen:
@sbillinge I've simplified the workflow using the idea from the diagram, please check.
In the mean time I'll think about how to make the tests more readable as also suggested by @bobleesj (since we'll add orcid this will make the tests extremely long to account for different cases).

yucongalicechen:
replaced by #267

PR #259, opened 2024-12-20
Remove 6 pytest warnings with capturing wavelength warnings

bobleesj:
Making smaller PRs since @sbillinge is on the road.

### DIFF
--- a/news/pytest-warnings-others.rst
+++ b/news/pytest-warnings-others.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* No news added
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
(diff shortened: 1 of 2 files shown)


news/pytest-warnings-others.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* No news added
```
sbillinge, line 3: please use something like:
```
* no news added: intercepting warnings generated by tests on purpose not needed by users or devs.
```
With that said, I think this probably would be good to hvae a news.....

sbillinge (commented):
code is good here so I will merge to keep things going, but please see the comments on the news.

news/pytest-warnings-others.rst
bobleesj: Yup noted. There are other remaining warnings, etc. I will include that news item

PR #264, opened 2024-12-21
refactor  to separate getting info and creating config files

### LINKED ISSUE #245: I think the `get_user_info` docs don't match the code

sbillinge:
closes #245 
replaces #253 

@yucongalicechen @bobleesj ready for review.  I will put the config-file creation workflow on another PR

Commented text will be reused (and removed from here) in a separate function for creating a missing global config file in a future PR.

### DIFF
--- a/doc/source/examples/tools_example.rst
+++ b/doc/source/examples/tools_example.rst
@@ -13,11 +13,11 @@ Automatically Capture User Info
 One task we would like to do is to capture and propagate useful metadata that describes the diffraction data.
 Some is essential such as wavelength and radiation type. Other metadata is useful such as information about the
 sample, co-workers and so on.  However, one of the most important bits of information is the name of the data owner.
-For example, in ``DiffractionObjects`` this is stored in the ``metadata`` dictionary as ``username``, ``user_email``,
-and ``user_orcid``.
+For example, in ``DiffractionObjects`` this is stored in the ``metadata`` dictionary as ``owner_name``, ``owner_email``,
+and ``owner_orcid``.
 
 To reduce experimenter overhead when collecting this information, we have developed an infrastructure that helps
-to capture this information automatically when you are using `DiffractionObjects` and other diffpy tools.
+to capture this information automatically when you are using ``DiffractionObjects`` and other diffpy tools.
 You may also reuse this infrastructure for your own projects using tools in this tutorial.
 
 This example will demonstrate how ``diffpy.utils`` allows us to conveniently load and manage user and package information.
@@ -28,8 +28,9 @@ Load user info into your program
 
 To use this functionality in your own code make use of the ``get_user_info`` function in
 ``diffpy.utils.tools`` which will search for information about the user, parse it, and return
-it in a dictionary object e.g. if the user is "Jane Doe" with email "janedoe@gmail.com" and the
-function can find the information, if you type this
+it in a dictionary object e.g. if the user is "Jane Doe" with email "janedoe@gmail.com" and ORCID
+"0000-0000-0000-0000", and if the
+function can find the information (more on this below), if you type this
 
 .. code-block:: python
 
@@ -40,7 +41,7 @@ The function will return
 
 .. code-block:: python
 
-        {"email": "janedoe@email.com", "username": "Jane Doe"}
+        {"owner_email": "janedoe@email.com", "owner_name": "Jane Doe", "owner_orcid": "0000-0000-0000-0000"}
 
 
 Where does ``get_user_info()`` get the user information from?
@@ -48,8 +49,9 @@ Where does ``get_user_info()`` get the user information from?
 
 The function will first attempt to load the information from configuration files with the name ``diffpyconfig.json``
 on your hard-drive.
-It looks first for the file in the current working directory.  If it cannot find it there it will look
-user's home, i.e., login, directory.  To find this directory, open a terminal and a unix or mac system type ::
+It looks for files in the current working directory and in the computer-user's home (i.e., login) directory.
+For example, it might be in C:/Users/yourname`` or something like that, but to find this directory, open
+a terminal and a unix or mac system type ::
 
     cd ~
     pwd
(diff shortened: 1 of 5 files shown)


doc/source/examples/tools_example.rst
```diff
@@ -40,16 +41,17 @@ The function will return
 
 .. code-block:: python
 
-        {"email": "janedoe@email.com", "username": "Jane Doe"}
+        {"owner_email": "janedoe@email.com", "owner_name": "Jane Doe", "owner_orcid": "0000-0000-0000-0000"}
```
bobleesj, line 44: i like the term "owner" - and orcid since that's a gateway for more info

doc/source/examples/tools_example.rst
```diff
@@ -40,16 +41,17 @@ The function will return
...
 
 Where does ``get_user_info()`` get the user information from?
 -------------------------------------------------------------
 
 The function will first attempt to load the information from configuration files with the name ``diffpyconfig.json``
 on your hard-drive.
-It looks first for the file in the current working directory.  If it cannot find it there it will look
-user's home, i.e., login, directory.  To find this directory, open a terminal and a unix or mac system type ::
+It looks for files in the current working directory and in the computer-user's home (i.e., login) directory.
+For example, it might be in C:/Users/yourname`` or something like that, but to find this directory, open
```
bobleesj, line 53: much clearer since there is an example provided since i wasn't sure what "logic" meant here.

doc/source/examples/tools_example.rst
```diff
@@ -58,67 +60,55 @@ Or type ``Echo $HOME``.  On a Windows computer ::
...
+        with open(filepath, "w") as f:
+            json.dump(metadata_enhanced_data, f)
+
+``check_and_build_global_config()`` only
+interrupts execution if it can't find a valid config file, and so if the user enters valid information
+it will only run once.  However, if you want to bypass this behavior,
+``check_and_build_global_config()`` takes an optional boolean ``skip_config_creation`` parameter that
+could be set to ``True`` at runtime to override the config creation.
+
+I entered the wrong information in my config file so it always loads incorrect information, how do I fix that?
```
bobleesj, line 98: Clear on how to reset config

src/diffpy/utils/tools.py
```diff
@@ -91,47 +90,62 @@ def _create_global_config(args):
     return return_bool
 
 
-def get_user_info(args=None):
+def get_user_info(owner_name=None, owner_email=None, owner_orcid=None):
```
bobleesj, line 93: much better - instead of `args`, this line explicitly communicates what info can be optionally provided

src/diffpy/utils/tools.py
```diff
@@ -91,47 +90,62 @@ def _create_global_config(args):
...
+      "owner_orcid": "<your_associated_orcid if you would like this stored with your data>>"
+    }
+    You may also store any other gloabl-level information that you would like associated with your
+    diffraction data in this file
 
     Parameters
     ----------
-    args argparse.Namespace
-        The arguments from the parser, default is None.
+    owner_name: string, optional, default is the value stored in the global or local config file.
```
bobleesj, line 115: Ah.. this gives me joy

tests/test_tools.py
```diff
@@ -84,44 +73,103 @@ def _run_tests(inputs, expected):
...
-    _run_tests(inputs, expected)
-
-
-@pytest.mark.parametrize("inputs, expected", params_user_info_with_local_conf_file)
-def test_get_user_info_with_local_conf_file(monkeypatch, inputs, expected, user_filesystem):
-    _setup_dirs(monkeypatch, user_filesystem)
-    local_config_data = {"username": "cwd_username", "email": "cwd@email.com"}
-    with open(Path(user_filesystem) / "diffpyconfig.json", "w") as f:
+@pytest.mark.parametrize(
+    "runtime_inputs, expected",
```
bobleesj, line 77: i like it - runtime is much clearer than `cli`

tests/test_tools.py
```diff
@@ -84,44 +73,103 @@ def _run_tests(inputs, expected):
 ]
 
 
-@pytest.mark.parametrize("inputs, expected", params_user_info_with_home_conf_file)
-def test_get_user_info_with_home_conf_file(monkeypatch, inputs, expected, user_filesystem):
-    _setup_dirs(monkeypatch, user_filesystem)
-    _run_tests(inputs, expected)
```
bobleesj, line 90: yes, this mysterious `_run_tests` func is replaced.

tests/test_tools.py
```diff
@@ -84,44 +73,103 @@ def _run_tests(inputs, expected):
...
+        ),
+        # C5: just owner ci passed in at runtime.  expect runtime_orcid but others from config
+        (
+            {"owner_orcid": "runtime_orcid"},
+            {"owner_name": "home_ownername", "owner_email": "home@email.com", "owner_orcid": "runtime_orcid"},
+        ),
+    ],
+)
+def test_get_user_info_with_home_conf_file(runtime_inputs, expected, user_filesystem, mocker):
+    # user_filesystem[0] is tmp_dir/home_dir with the global config file in it, user_filesystem[1]
```
bobleesj, line 104: i was wondering `user_filesystem` and the comment clarifies it

bobleesj:
@sbillinge I did my review via in-line comments. 

One general question, should we follow pep8 convention for comments? 1) capitalize the first word 2) use complete sentences?

> PEP8 Comments should be complete sentences. The first word should be capitalized, unless it is an identifier that begins with a lower case letter (never alter the case of identifiers!).

REF https://peps.python.org/pep-0008/#comments

sbillinge:
@yucongalicechen @bobleesj I need the updated `user_filesystem` for the other PR on the config updater so I will merge this to get it.  i think it is in pretty good shape (thanks @bobleesj for the feedback).  @yucongalicechen please take a look at this PR when you get a chance and if there is anything you find needs changing we can do it on another PR.

src/diffpy/utils/tools.py
```diff
@@ -65,7 +64,7 @@ def load_config(file_path):
             config = json.load(f)
         return config
     else:
-        return None
+        return {}
 
 
 def _sorted_merge(*dicts):
```
yucongalicechen, line 70: I wonder do we want to make some of the functions private (like `load_config`)?

src/diffpy/utils/tools.py
```diff
@@ -91,47 +90,62 @@ def _create_global_config(args):
...
 
     Parameters
     ----------
-    args argparse.Namespace
-        The arguments from the parser, default is None.
+    owner_name: string, optional, default is the value stored in the global or local config file.
+        The name of the user who will show as owner in the metadata that is stored with the data
+    owner_email: string, optional, default is the value stored in the global or local config file.
+        The email of the user/owner
+    owner_name:  string, optional, default is the value stored in the global or local config file.
```
yucongalicechen, line 119: typo - owner_orcid

doc/source/examples/tools_example.rst
```diff
@@ -58,67 +60,55 @@ Or type ``Echo $HOME``.  On a Windows computer ::
...
+--------------------------------------------------------------------------------------------------------------
+
+It is easy to fix this simply by deleting the global and/or local config files, which will allow
+you to re-enter the information during the ``check_and_build_global_config()`` initialization
+workflow.   You can also simply editi the ``diffpyconfig.json`` file directly using a text
```
yucongalicechen, line 103: typo - edit

yucongalicechen:
I finished my review @sbillinge here. Like this new structure and just caught two typos.

src/diffpy/utils/tools.py
sbillinge: yes, this should probably be private, I agree.

Can we also check?  I don't think we need `_sorted_dict()` any more so that could be deleted on a new PR.

src/diffpy/utils/tools.py
sbillinge: thanks, good catch.

src/diffpy/utils/tools.py
yucongalicechen: Yes, I can fix these on a new PR, once that updater workflow PR is merged

sbillinge:
Thanks for the excellent review @yucongalicechen .  Please could you make a PR to fix those things?  Let's also double check there are no private functions that are not use, or public functions that we want to make private.

If @bobleesj finds anything else we can put that into the same PR.

I also wanted to add into the docs what happens in the update config workflow when it is triggered, so reproduce what the user would see and give an example of how they would respond.  It would be great for me if that could be in the same PR?

yucongalicechen:
> Thanks for the excellent review @yucongalicechen . Please could you make a PR to fix those things? Let's also double check there are no private functions that are not use, or public functions that we want to make private.
> 
> If @bobleesj finds anything else we can put that into the same PR.
> 
> I also wanted to add into the docs what happens in the update config workflow when it is triggered, so reproduce what the user would see and give an example of how they would respond. It would be great for me if that could be in the same PR?

Sounds good, I'll do that

PR #266, opened 2024-12-22
Continue refactor test functions, replace `UC` to `C` in `test_transforms.py`

bobleesj:


tests/test_diffraction_objects.py
bobleesj, line 1: In the group, do we want to call it "C" instead of "Case"?

tests/test_transforms.py
```diff
@@ -39,13 +40,14 @@ def test_q_to_tth(wavelength, q, expected_tth, wavelength_warning_msg):
...
     [
-        # UC1: user specified invalid q values that result in tth > 180 degrees
+        # Test error messages in q to tth conversion with invalid Two theta values.
+        # C1: Invalid q values that result in tth > 180 degrees
         (
             4 * np.pi,
             np.array([0.2, 0.4, 0.6, 0.8, 1, 1.2]),
             ValueError,
         ),
-        # UC2: user specified a wrong wavelength that result in tth > 180 degrees
```
bobleesj, line 48: a comment starting with `user specified...` feels a bit repetitive to me and does not add great value..

tests/test_transforms.py
```diff
@@ -62,17 +64,17 @@ def test_q_to_tth_bad(wavelength, q, expected_error_type, invalid_q_or_d_or_wave
...
         (4 * np.pi, np.array([]), np.array([])),
-        # UC2: user specified valid tth values between 0-180 degrees (without wavelength)
+        # C2: valid tth values between 0-180 degrees (without wavelength)
         (
             None,
             np.array([0, 30, 60, 90, 120, 180]),
             np.array([0, 1, 2, 3, 4, 5]),
         ),
-        # UC3: user specified valid tth values between 0-180 degrees (with wavelength)
+        # C3: valid tth values between 0-180 degrees (with wavelength)
```
bobleesj, line 77: I don't think we want to call it `UC` right? Since this is testing `actual_q = tth_to_q(tth, wavelength)`

tests/test_transforms.py
```diff
@@ -94,14 +96,14 @@ def test_tth_to_q(wavelength, tth, expected_q, wavelength_warning_msg):
 @pytest.mark.parametrize(
     "wavelength, tth, expected_error_type, expected_error_msg",
     [
-        # UC0: user specified an invalid tth value of > 180 degrees (without wavelength)
+        # C1: invalid tth value of > 180 degrees, no wavelength, expect two theta ValueError
```
bobleesj, line 99: same comment as above - testing `tth_to_q(tth, wavelength)`

tests/test_transforms.py
```diff
@@ -171,13 +173,13 @@ def test_d_to_q(d, expected_q, zero_divide_error_expected):
     "wavelength, tth, expected_d, divide_by_zero_warning_expected",
     [
         # Test conversion of q to d with valid values
-        # Case 1: empty tth values, no, expect empty d values
+        # C1: empty tth values, no, expect empty d values
```
bobleesj, line 176: same comment - testing `tth_to_d`

tests/test_diffraction_objects.py
```diff
@@ -186,7 +186,7 @@ def test_init_invalid_xtype():
     "org_do_args, target_do_args, scale_inputs, expected",
     [
         # Test that scale_to() scales to the correct values
-        # Case 1: same x-array and y-array, check offset
+        # C1: same x-array and y-array, check offset
```
bobleesj, line 189: These are test sets for the `scale_to` method

tests/test_diffraction_objects.py
```diff
@@ -287,76 +287,81 @@ def test_scale_to(org_do_args, target_do_args, scale_inputs, expected):
...
+        original_do.scale_to(
+            target_do,
+            q=scale_inputs["q"],
+            tth=scale_inputs["tth"],
+            d=scale_inputs["d"],
+            offset=scale_inputs["offset"],
         )
 
 
-params_index = [
```
bobleesj, line 347: just moving `params_index` under `@pytest.mark.parametrize`

tests/test_diffraction_objects.py
```diff
@@ -406,7 +411,7 @@ def test_dump(tmp_path, mocker):
 @pytest.mark.parametrize(
     "do_init_args, expected_do_dict, divide_by_zero_warning_expected",
     [
-        (  # instantiate just array attributes
+        (  # Instantiate just array attributes
```
bobleesj, line 414: Quick PEP8 standard, starting each comment with a cap letter

tests/test_transforms.py
```diff
@@ -9,20 +9,20 @@
 @pytest.mark.parametrize(
     "wavelength, q, expected_tth",
     [
-        # Case 1: Allow empty arrays for q
-        # 1. Empty q values, no wavelength, return empty arrays
-        (None, np.empty((0)), np.empty((0))),
-        # 2. Empty q values, wavelength specified, return empty arrays
+        # Test conversion of q to tth with valid values
+        # C1: Allow empty array q to compute tth with or without wavelength
+        # 1. Wavelength provided, expect empty array of tth
```
bobleesj, line 14: I think the word `expect` is better than `return` since expect can be used for warnings too.

Would you have any further feedback on this refactored version?

bobleesj: The flow is like..

"The overall goal is  to`Test conversion of q to tth with valid values`. For the first case, we want to check whether `Allow empty array q to compute tth with or without wavelength`. Then, for the first case, when `Wavelength provided, expect empty array of tth`. Now, When wavelength, expect you would get an empty tth value.. Now, when `No wavelength provided`, still expect an empty tth but hey you will also get a warning!"

tests/test_transforms.py
```diff
@@ -94,14 +95,14 @@ def test_tth_to_q(wavelength, tth, expected_q, wavelength_warning_msg):
 @pytest.mark.parametrize(
     "wavelength, tth, expected_error_type, expected_error_msg",
     [
-        # UC0: user specified an invalid tth value of > 180 degrees (without wavelength)
+        # C1: Invalid tth value of > 180 degrees, no wavelength, expect two theta ValueError
```
bobleesj: More refinements needed for the remaining in separate PRs.

bobleesj:
@sbillinge ready for review

tests/test_diffraction_objects.py
sbillinge: I think either can work

tests/test_transforms.py
sbillinge: We basically never want to see UCs in tests.  UCs are enumerated lists of actions that describe user behavior when using an app.

tests/test_transforms.py
sbillinge: Right, no UC

tests/test_transforms.py
sbillinge: Either is fine

tests/test_diffraction_objects.py
sbillinge: Yes, let's use that convention.  For the test cases I don't mind if they are full sentences or not tbh as long as the intent is clear.

tests/test_transforms.py
sbillinge: Your version looks great here
 Very clear

tests/test_diffraction_objects.py
```diff
@@ -287,76 +287,81 @@ def test_scale_to(org_do_args, target_do_args, scale_inputs, expected):
...
-    target_diff_object = DiffractionObject(
-        xarray=inputs["target_xarray"],
-        yarray=inputs["target_yarray"],
-        xtype=inputs["target_xtype"],
-        wavelength=inputs["target_wavelength"],
-    )
+@pytest.mark.parametrize(
+    "org_do_args, target_do_args, scale_inputs",
+    [
+        # UC1: User did not specify anything
```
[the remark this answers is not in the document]
sbillinge: Wait, I read above... You will do these on separate PRs, sorry!

sbillinge (commented):
Please see inline. It looks great. I will go ahead and merge.

bobleesj:
> Yes, let's use that convention. For the test cases I don't mind if they are full sentences or not tbh as long as the intent is clear.

Right, method docstrings also doesn't have to start with a complete "noun-verb" sentences anyway.. so i agree.

PR #267, opened 2024-12-22
Config updater workflow

### LINKED ISSUE #244: implement a `skip_config_creation` option in `get_user_info`

sbillinge:
closes #244

This is the new separate workflow for updating the config file if it is found to be missing in home.


sbillinge:
@bobleesj @yucongalicechen @alisnwu this is ready for review

src/diffpy/utils/tools.py
```diff
@@ -132,22 +116,50 @@ def get_user_info(owner_name=None, owner_email=None, owner_orcid=None):
...
-    #         "No global configuration file was found containing "
-    #         "information about the user to associate with the data.\n"
-    #         "By following the prompts below you can add your name and email to this file on the current "
-    #         "computer and your name will be automatically associated with subsequent diffpy data by default.\n"
-    #         "This is not recommended on a shared or public computer. "
-    #         "You will only have to do that once.\n"
-    #         "For more information, please refer to www.diffpy.org/diffpy.utils/examples/toolsexample.html"
-    #     )
     user_info = global_config
     user_info.update(local_config)
```
yucongalicechen, line 120: Since user can only create/update global config, I wonder if it is better to prioritize global to local config?

sbillinge: @yucongalicechen thanks for the comment.  User updates local config by placing a file in their local directory.  This is a pretty standard protocol....local overrides global.

yucongalicechen: Gotcha. Then this PR looks great to me. I like the separated structure a lot.

sbillinge: Thanks @yucongalicechen !  Yes, it is something I learned at some point.  When you are struggling with function logic it is often because the function is trying to do more than one thing and can be simplified by making two functions that each do one thing.

src/diffpy/utils/tools.py
```diff
@@ -132,22 +116,50 @@ def get_user_info(owner_name=None, owner_email=None, owner_orcid=None):
...
+            f.write(json.dumps(config))
+        outro_text = (
+            f"The config file at {Path().home() / 'diffpyconfig.json'} has been created. "
+            f"The values  {config} were entered.\n"
+            f"These values will be inserted as metadata with your data in apps that use "
+            f"diffpy.get_user_info(). If you would like to update these values, either "
+            f"delete the config file and this workflow will rerun next time you run this "
+            f"program.  Or you may open the config file in a text editor and manually edit the"
+            f"entries.  For more information, see: "
+            f"https://diffpy.githu.io/diffpy.utils/examples/tools_example.html"
```
yucongalicechen, line 157: github - typo

yucongalicechen:
Just another comment - looks like this function does not have docstring yet

sbillinge (commented):
Thanks for the comments @yucongalicechen.  I fixed them.  Writing the docstring also made me realize we may want to return the status of if a config file is found.  Then a user could use the function with the skip but if it returns False make their own workflow, which is nice and flexible.   Great reviews everyone!

bobleesj:
seems like a conflict created now hm..

sbillinge:
Ok, I will fix and merge

Issue #269, opened 2024-12-22
Do we want to call the DO unique id as `id` or `uuid`?


bobleesj:
_Originally posted by @sbillinge in https://github.com/diffpy/diffpy.utils/pull/265#discussion_r1894953424_

I like `uuid` because it is not only specific but also it implies it's a variable/constant that the user should not play with.

sbillinge:
I like the discussion. How about `_uuid`, which is even more hands off.

bobleesj:
right, `_uuid` I like it - we (maintainers) will be the primary user of this variable. 

For regular users, they could use the "name" mutable variable, written in human language.

bobleesj:
Will create a quick PR for this

PR #271, opened 2024-12-23
Rename `id` to `uuid` in DiffractionObject

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/269

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -288,9 +288,9 @@ class DiffractionObject:
         Examples
         --------
         >>> my_do.all_arrays[:, 0]  # yarray
-        >>> my_do.all_arrays[:, 1]  # `xarray` in q
-        >>> my_do.all_arrays[:, 2]  # `xarray` in tth
-        >>> my_do.all_arrays[:, 3]  # `xarray` in d
+        >>> my_do.all_arrays[:, 1]  # xarray in q
+        >>> my_do.all_arrays[:, 2]  # xarray in tth
+        >>> my_do.all_arrays[:, 3]  # xarray in d
         """
         return self._all_arrays
 
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -50,7 +50,7 @@ class DiffractionObject:
         The array containing the quantity of q, tth, d values.
     input_xtype : str
         The type of the independent variable in `xarray`. Must be one of {*XQUANTITIES}
-    id : uuid
+    _uuid : uuid
         The unique identifier for the diffraction object.
     scat_quantity : str
         The type of scattering experiment (e.g., "x-ray", "neutron"). Default is an empty string "".
@@ -127,7 +127,7 @@ class DiffractionObject:
         >>> print(do.metadata)
         """
 
-        self._id = uuid.uuid4()
+        self._uuid = uuid.uuid4()
         self._input_data(xarray, yarray, xtype, wavelength, scat_quantity, name, metadata)
 
     def _input_data(self, xarray, yarray, xtype, wavelength, scat_quantity, name, metadata):
@@ -299,12 +299,12 @@ class DiffractionObject:
         raise AttributeError(_setter_wmsg("input_xtype"))
 
     @property
-    def id(self):
-        return self._id
+    def uuid(self):
+        return self._uuid
 
-    @id.setter
-    def id(self, _):
-        raise AttributeError(_setter_wmsg("id"))
+    @uuid.setter
+    def uuid(self, _):
+        raise AttributeError(_setter_wmsg("uuid"))
 
     def get_array_index(self, value, xtype=None):
         """Return the index of the closest value in the array associated with
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -284,12 +284,12 @@ class DiffractionObject:
         -------
         ndarray
             The shape (len(data), 4) 2D array with columns containing the `yarray` (intensity)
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -299,12 +299,12 @@ def input_xtype(self, _):
...
     @property
-    def id(self):
-        return self._id
+    def uuid(self):
+        return self._uuid
 
-    @id.setter
-    def id(self, _):
-        raise AttributeError(_setter_wmsg("id"))
+    @uuid.setter
```
bobleesj, line 305: while User still has access to `uuid`, we make it non-settable still.

bobleesj (commented):
@sbillinge ready for review

src/diffpy/utils/diffraction_objects.py
```diff
@@ -50,7 +50,7 @@ class DiffractionObject:
         The array containing the quantity of q, tth, d values.
     input_xtype : str
         The type of the independent variable in `xarray`. Must be one of {*XQUANTITIES}
-    id : uuid
+    _uuid : uuid
```
sbillinge, line 53: if we are making this attribute public and giving it a getter, I wonder if we should just call it uuid and remove the underscore?  What do you think?  It may also be less confusing if the attribute name and how it is accessed is the szme?

sbillinge (commented):
just one thought.

src/diffpy/utils/diffraction_objects.py
bobleesj: Yes, I agree. It turns out we shall write a separate docstring for `@property` methods as shown below (as done in the latest commit):

```
    @property
    def uuid(self):
        """The unique identifier for the DiffractionObject instance.

        Returns
        -------
        uuid
            The unique identifier of the DiffractionObject instance.
        """
        return self._uuid
```

This way, User does not have to read private properties starting with `_`.

API doc is also clearly marked with `@property`:

Screenshot 2024-12-25 at 11 12 35 PM

src/diffpy/utils/diffraction_objects.py
```diff
@@ -46,12 +46,6 @@ class DiffractionObject:
 
     Attributes
     ----------
-    all_arrays : ndarray
-        The array containing the quantity of q, tth, d values.
-    input_xtype : str
-        The type of the independent variable in `xarray`. Must be one of {*XQUANTITIES}
-    id : uuid
-        The unique identifier for the diffraction object.
```
bobleesj, line 54: These are marked with `@property` - docstrings are provided separately.

bobleesj:
@sbillinge thanks for your review - ready for review, with my reply above: https://github.com/diffpy/diffpy.utils/pull/271#discussion_r1897608557

src/diffpy/utils/diffraction_objects.py
```diff
@@ -284,6 +278,20 @@ def __rtruediv__(self, other):
 
     @property
     def all_arrays(self):
+        """The 2D array containing `xarray` and `yarray` values.
+
+        Returns
+        -------
+        ndarray
+            The 2D array containing the `xarray` values in q, tth, d, and `yarray`.
```
sbillinge, line 286: for lower cognitive load I would change to `The `shape(len(data), 4)` 2D array with columns containing the `yarray` (intensity) and the `xarray` values in q, tth, and d .` so they appear in the docstring in the same order as in the array columns

src/diffpy/utils/diffraction_objects.py
```diff
@@ -284,6 +278,20 @@ def __rtruediv__(self, other):
...
+
+        Examples
+        --------
+        >>> my_do.all_arrays[:, 0]  # yarray
```
sbillinge, line 290: add "To access specific arrays individually, use these slices:"?

src/diffpy/utils/diffraction_objects.py
```diff
@@ -319,7 +341,8 @@ def get_array_index(self, value, xtype=None):
 
         Returns
         -------
-        the index of the value in the array
+        list
```
sbillinge, line 344: let's check.  Logically, this should return an int not a list.  Let's check and maybe update the function/tesnts, or understand why it returns a list and change the name.

sbillinge (commented):
few questions inline.

src/diffpy/utils/diffraction_objects.py
bobleesj: Yes, great! superb. The example and this docstring makes is super clear.

src/diffpy/utils/diffraction_objects.py
bobleesj: Screenshot 2024-12-26 at 2 46 56 PM

Yes, great. done

src/diffpy/utils/diffraction_objects.py
bobleesj: Yup, I created a separate issue for this since it's beyond the scope of this PR. I will look into this.

bobleesj:
@sbillinge ready for review - addressed your comments. thanks for the review

src/diffpy/utils/diffraction_objects.py
bobleesj: https://github.com/diffpy/diffpy.utils/issues/279

src/diffpy/utils/diffraction_objects.py
sbillinge: There's a stray y-array at the end of the line :(

bobleesj: Sorry - fixed now:        

---

            The shape (len(data), 4) 2D array with columns containing the `yarray` (intensity)
            and the `xarray` values in q, tth, and d.

bobleesj:
@sbillinge ready for review - fixed the extra yarray

PR #272, opened 2024-12-23
Apply `docformatter` in `pre-commit` for PEP 257 docstring auto format, no manual modifications made

bobleesj:
Just ran `pre-commit` only - I think it's quite useful


src/diffpy/utils/resampler.py
```diff
@@ -80,7 +79,8 @@ def wsinterp(x, xp, fp, left=None, right=None):
 
 
 def nsinterp(xp, fp, qmin=0, qmax=25, left=None, right=None):
-    """One-dimensional Whittaker-Shannon interpolation onto the Nyquist-Shannon grid.
+    """One-dimensional Whittaker-Shannon interpolation onto the Nyquist-Shannon
```
bobleesj, line 82: nice - it automatically makes the first line within a certain line length.

tests/test_loaddata.py
```diff
@@ -10,7 +9,7 @@
 
 
 def test_loadData_default(datafile):
-    """check loadData() with default options"""
+    """Check loadData() with default options."""
```
bobleesj, line 12: Nice - automatic start with cap letter

bobleesj:
@sbillinge I think it's super useful.. I could add this to our cookiecutting standard if you find it useful too.

sbillinge:
@bobleesj this is conflicted now.  I think it would be a good idea to wait until #267 is merged (I think that is the right one, the config creator) and then merge main into this and run it again.  Otherwise this will make that one conflicted.

bobleesj:
> @bobleesj this is conflicted now. 

Yes, will wait for the user config docstrings to be merged first.

bobleesj:
@sbillinge ready for review - resolved the conflict

PR #273, opened 2024-12-23
Improve test comments under `@pytest.mark.parametrize` in `test_diffraction_objects.py`

bobleesj:
Some battle-tested practices being adopted:

- Replace UC to C
- Add higher-level test function comment before starting "C"
- Use "x, y, z (provided), expect blah blah blah" format for consistency

cc' @yucongalicechen

### DIFF
--- a/tests/test_diffraction_objects.py
+++ b/tests/test_diffraction_objects.py
@@ -12,10 +12,10 @@ from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
 
 
 @pytest.mark.parametrize(
-    "do_args_1, do_args_2, expected_equality, warning_expected",
+    "do_args_1, do_args_2, expected_equality, wavelength_warning_expected",
     [
         # Test when __eq__ returns True and False
-        # Identical args, expect equality
+        # C1: Identical args, expect equality
         (
             {
                 "name": "same",
@@ -56,7 +56,7 @@ from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
             False,
             True,
         ),
-        (  # One without wavelength, expect inequality
+        (  # C2: One without wavelength, expect inequality
             {
                 "wavelength": 0.71,
                 "xtype": "tth",
@@ -73,7 +73,7 @@ from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
             False,
             True,
         ),
-        (  # Different wavelengths, expect inequality
+        (  # C3: Different wavelength values, expect inequality
             {
                 "wavelength": 0.71,
                 "xtype": "tth",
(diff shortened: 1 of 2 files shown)


bobleesj:
@sbillinge ready for review.

@yucongalicechen , the part that is remaining is the `test_scale_to` function. 

The existing comments  like`# C2: Same length x-arrays with exact x-value match`, `# C3: Same length x-arrays with approximate x-value match` were abit hard for me to understand what's being tested. Would you mind improving this test comments for `test_scale_to` using the above format?

bobleesj:
(Just added a quick fix) pls see my comments right above

tests/test_diffraction_objects.py
```diff
@@ -350,21 +363,84 @@ def test_scale_to_bad(org_do_args, target_do_args, scale_inputs):
...
-        # UC1: Exact match
-        (4 * np.pi, np.array([30.005, 60]), np.array([1, 2]), "tth", "tth", 30.005, [0]),
-        # UC2: Target value lies in the array, returns the (first) closest index
-        (4 * np.pi, np.array([30, 60]), np.array([1, 2]), "tth", "tth", 45, [0]),
-        (4 * np.pi, np.array([30, 60]), np.array([1, 2]), "tth", "q", 0.25, [0]),
-        # UC3: Target value out of the range, returns the closest index
-        (4 * np.pi, np.array([0.25, 0.5, 0.71]), np.array([1, 2, 3]), "q", "q", 0.1, [0]),
-        (4 * np.pi, np.array([30, 60]), np.array([1, 2]), "tth", "tth", 63, [1]),
+        # Test get_array_index() returns the expected index given xtype and value
+        # C1: Target value is in the xarray and xtype is identical, expect exact index match
```
sbillinge, line 369: let's not modify this PR again, but I do wonder whether it would be more compact as
```
        (  # C1: Target value is in the xarray and xtype is identical, expect exact index match
            {
                "wavelength": 4 * np.pi,
                "xarray": np.array([30.005, 60]),
                "yarray": np.array([1, 2]),
                "xtype": "tth",
            },
            {
                "xtype": "tth",
                "value": 30.005,
            },
            [0],
        ),
```
than the current
```
    # C1: Target value is in the xarray and xtype is identical, expect exact index match
        (
            {
                "wavelength": 4 * np.pi,
                "xarray": np.array([30.005, 60]),
                "yarray": np.array([1, 2]),
                "xtype": "tth",
            },
            {
                "xtype": "tth",
                "value": 30.005,
            },
            [0],
        ),
```

sbillinge (commented):
please see my comment.  I guess this may not be liked by one of our linters or something, but I think I prefer the more compact formatting.  Not a huge deal though.  Let's just decide which we like and adopt it in future.  It is such a small point we don't need to redo things here.  Let's dicsus it.  But I will merge this.

tests/test_diffraction_objects.py
bobleesj: Screenshot 2024-12-25 at 10 07 41 AM

@sbillinge yeah I like that! The automatic indentation also provides a good visual hierarchy!

bobleesj: Screenshot 2024-12-25 at 10 11 26 AM

For single-lined cases, I think we still want to maintain this format right?

PR #278, opened 2024-12-26
compute mu

### LINKED ISSUE #45: mu T calculator
This is functionality that @yucongalicechen is working on for `labpdfproc`.  The UC is that the user inputs some basic information about the sample and experiment (composition and wavelength) and the program returns a mu (linear absorption coefficient) for the sample.

The idea is to put this functionality into `diffpy.utils` instead so it could be used by other applications.

Probably in diffpy.utils we want it as a library capability (i.e., not an app with an argparser)  the arg.parser would be in labpdfproc, this would just have the lookup tables and the functionality.

It is possible that some of the functionality (if not all) is available elsewhere in which case we could decide whether to include that as a dependency instead of implementing it ourselves.

yucongalicechen:
closes #45 
@sbillinge ready for review. I can add docs in the same PR once we're happy with the function


src/diffpy/utils/tools.py
```diff
@@ -131,3 +133,25 @@ def get_package_info(package_names, metadata=None):
...
+
+
+def compute_mu_using_xraydb(sample, energy, density=None):
+    """Compute mu using the XrayDB database.
+
+    Reference: https://xraypy.github.io/XrayDB/python.html#xraydb.material_mu
+
+    Parameters
+    ----------
+    sample str
```
sbillinge, line 145: I think we would normally do tihs as:
```
    sample: str
       The chemical formula or the name of the material
```
i.e., colon and sentencecse

src/diffpy/utils/tools.py
```diff
@@ -131,3 +133,25 @@ def get_package_info(package_names, metadata=None):
...
+        the chemical formula or the name of the material
+    energy float
+        the energy in eV
+    density float or None
```
sbillinge, line 149: `density: float, optional, Default is None`

src/diffpy/utils/tools.py
```diff
@@ -131,3 +133,25 @@ def get_package_info(package_names, metadata=None):
...
+        material density in gr/cm^3
```
sbillinge, line 150: this should be "sample mass density" or even better, "mass density of the packed powder/sample" to be clear.  the material desnsity is something quite different (mass of atoms in the unit cell divided by unit cell volume)

src/diffpy/utils/tools.py
```diff
@@ -131,3 +133,25 @@ def get_package_info(package_names, metadata=None):
         pkg_info.update({package: importlib.metadata.version(package)})
     metadata["package_info"] = pkg_info
     return metadata
+
+
+def compute_mu_using_xraydb(sample, energy, density=None):
```
sbillinge, line 138: let's change `sample` to `composition` or maybe `sample_composition`.

I think we probably  need `packing_fraction` as an optional arg.

tests/test_tools.py
```diff
@@ -189,3 +189,9 @@ def test_get_package_info(monkeypatch, inputs, expected):
     )
     actual_metadata = get_package_info(inputs[0], metadata=inputs[1])
     assert actual_metadata == expected
+
+
+def test_compute_mu_using_xraydb():
```
sbillinge, line 194: we need a test for no density passed in.  Also for packing fraction passed in and also for both passed in.  Think about what behavior you want in each of these cases and write a "case" (look elsewhere in test_tools for how we are handling this now)

sbillinge (commented):
please see inline

yucongalicechen:
@sbillinge ready for another review

src/diffpy/utils/tools.py
```diff
@@ -131,3 +133,31 @@ def get_package_info(package_names, metadata=None):
         pkg_info.update({package: importlib.metadata.version(package)})
     metadata["package_info"] = pkg_info
     return metadata
+
+
+def compute_mu_using_xraydb(sample_composition, energy, density=None, packing_fraction=1):
```
yucongalicechen, line 138: added packing fraction. does this make sense or should we use None instead? I think if we set the default to 1 the code can be more concise

tests/test_tools.py
```diff
@@ -189,3 +189,27 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+params_mu = [
+    # C1: user didn't specify density or packing fraction
+    ({"sample_composition": "H2O", "energy": 10000, "density": None, "packing_fraction": 1}, 0.5330),
+    # C2: user specified packing fraction only
+    ({"sample_composition": "H2O", "energy": 10000, "density": None, "packing_fraction": 0.5}, 0.2665),
+    # C3: user specified density only
+    ({"sample_composition": "H2O", "energy": 10000, "density": 0.997, "packing_fraction": 1}, 0.5330),
+    ({"sample_composition": "H2O", "energy": 10000, "density": 0.4985, "packing_fraction": 1}, 0.2665),
+    # C4: user specified a standard density and a packing fraction
+    ({"sample_composition": "H2O", "energy": 10000, "density": 0.997, "packing_fraction": 0.5}, 0.2665),
```
yucongalicechen, line 203: added test cases with description. I didn't include the bad cases where the user does not specify a density and xraydb doesn't know the density (this is taken care by xraydb)

tests/test_tools.py
```diff
@@ -191,7 +191,25 @@ def test_get_package_info(monkeypatch, inputs, expected):
     assert actual_metadata == expected
 
 
-def test_compute_mu_using_xraydb():
-    sample, energy, density = "ZrO2", 17445.362740959618, 1.009
-    actual_mu = compute_mu_using_xraydb(sample, energy, density=density)
-    assert actual_mu == pytest.approx(1.252, rel=1e-4, abs=0.1)
+params_mu = [
```
bobleesj, line 194: Could you please follow the guides provided here? https://gitlab.thebillingegroup.com/resources/group-wiki/-/wikis/Group-standards-for-pytest-and-docstrings

add higher level function, use descriptive variable names, no need to create extra variable of `params_mu`,  no need to add "user specific, etc."

bobleesj: @yucongalicechen 

@sbillinge if you have also further suggestions to the gitlab doc, please suggest or modify directly

tests/test_tools.py
```diff
@@ -189,3 +189,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
     )
     actual_metadata = get_package_info(inputs[0], metadata=inputs[1])
     assert actual_metadata == expected
+
+
+@pytest.mark.parametrize(
+    "inputs, expected_mu",
+    [
+        # Test whether the function returns the correct mu
```
yucongalicechen, line 197: reorganized test comments

yucongalicechen:
thanks @bobleesj the link is very helpful and very clear to me. I edited the test comments. let me know if anything is unclear!
@sbillinge now ready for another review

sbillinge (commented):
please see inline comments.  ATM I am just reviewing tests until we get them sorted.  I don't really know what "standard density", and it doesn't make too much sense to me, so could you elaborate on that?

tests/test_tools.py
```diff
@@ -189,3 +189,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+        (  # C1: No density or packing fraction provided, expect to compute mu based on standard density
```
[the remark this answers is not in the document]
yucongalicechen: Just density I think - density under the standard conditions. So user can specify either a measured density or packing fraction.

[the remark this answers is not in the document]
yucongalicechen: The reason I'm not putting a strict "and/or" for density and packing fraction is that xraydb does not know the density to some materials like ZrO2. In this case user needs to put a density and a packing fraction, or they can put their measured density.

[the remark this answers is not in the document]
sbillinge: I see, so density is in the database?  that is unexpected.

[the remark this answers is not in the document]
sbillinge: how does xraydb compute the density?

tests/test_tools.py
sbillinge: gitlab doc looks pretty good to me.  Lot's of great examples to work from.

tests/test_tools.py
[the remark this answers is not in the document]
yucongalicechen: They have densities for some materials (for example, this returns sample name and density: https://xraypy.github.io/XrayDB/python.html#xraydb.get_materials).
I just reread their docstring and their density refers to the material density, not mass density: https://xraypy.github.io/XrayDB/python.html#xraydb.material_mu.

sbillinge (commented):
Please see inline comments. 

Also, this is conflicted

yucongalicechen:
> Please see inline comments.
> 
> Also, this is conflicted

So I think it only has density of pure elements and common compounds. I used `get_materials` to print out the list - it's pretty short.

I read through `material_mu` here: https://github.com/xraypy/XrayDB/blob/master/python/xraydb/materials.py. It uses a weighted sum approach for unknown materials. Here's the related code:

mass_tot, mu = 0.0, 0.0
    for elem, frac in chemparse(formula).items():
        mass  = frac * atomic_mass(elem)
        mu   += mass * mu_elam(elem, energy, kind=kind) # mu_elam is the mu for the given element
        mass_tot += mass
    return density*mu/mass_tot

It looks like it's using mass density so we're good. I'll solve the conflict and add more tests for unknown materials.

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,31 @@ def get_package_info(package_names, metadata=None):
...
+    User can provide a measured density or an estimated packing fraction.
+    Specifying the density is recommended, though not required for some pure or standard materials.
+    Reference: https://xraypy.github.io/XrayDB/python.html#xraydb.material_mu.
+
+    Parameters
+    ----------
+    sample_composition : str
+        The chemical formula or the name of the material.
+    energy : float
+        The energy in keV.
```
yucongalicechen, line 226: change unit to keV

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,31 @@ def get_package_info(package_names, metadata=None):
...
     metadata["package_info"] = pkg_info
     return metadata
+
+
+def compute_mu_using_xraydb(sample_composition, energy, density=None, packing_fraction=1):
+    """Compute the attenuation coefficient (mu) using the XrayDB database.
+
+    Computes mu based on the sample composition and energy.
+    User can provide a measured density or an estimated packing fraction.
+    Specifying the density is recommended, though not required for some pure or standard materials.
```
yucongalicechen, line 218: edit docstring

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+        (  # C3: Density provided only, expect to compute mu based on it
+            # 1. Known material
+            {
+                "sample_composition": "H2O",
+                "energy": 10,
+                "density": 0.987,
+            },
+            0.5330,
+        ),
+        (  # 2. Unknown material
```
yucongalicechen, line 198: add unknown material for testing

yucongalicechen:
@sbillinge ready for another review

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+                "energy": 17,
+                "density": 1.009,
+                "packing_fraction": 0.5,
+            },
+            0.626,
+        ),
+    ],
+)
+def test_compute_mu_using_xraydb(inputs, expected_mu):
+    actual_mu = compute_mu_using_xraydb(**inputs)
```
bobleesj, line 228: this is beautiful. thanks! @yucongalicechen

src/diffpy/utils/tools.py
sbillinge: I think setting to None is clearer here.   The logic from the perspective of the user is that we either have a packing fraction or we don't.  If it is None, you can set it to 1 in the code.

sbillinge: Also, didn't we discuss to change the name to `sample_mass_density`?

src/diffpy/utils/tools.py
sbillinge: I think using xrayDB functionality here is not useful.  The number of materials is so small, so there is no point to offer this to users as an option.  It will just confuse them.

We could offer other approaches:
1) given a composition, try and get density from COD or Materials Project.  For different polymorphs the densities will be different for the same composition, so we would have to handle that.
2) as a fallback, use a "best effort" atomic density (most tightly packed materials are around 5x10^22 cm^-3)?  This would be "magic" which we don't normally like, but if we use this fallback we could give a warning.

src/diffpy/utils/tools.py
sbillinge: "The energy of the incident x-rays in keV"

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,31 @@ def get_package_info(package_names, metadata=None):
...
+    Reference: https://xraypy.github.io/XrayDB/python.html#xraydb.material_mu.
+
+    Parameters
+    ----------
+    sample_composition : str
+        The chemical formula or the name of the material.
+    energy : float
+        The energy in keV.
+    density : float, optional, Default is None
+        The mass density of the packed powder/sample in gr/cm^3.
```
sbillinge, line 228: we need more info here about how these will be used, and what happens if someone specifies both.

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
     assert actual_metadata == expected
+
+
+@pytest.mark.parametrize(
+    "inputs, expected_mu",
+    [
+        # Test whether the function returns the correct mu
+        (  # C1: No density or packing fraction (only for known material), expect to get mu from database
+            {
+                "sample_composition": "H2O",
```
sbillinge, line 176: water is probably not a good "illustrative" choice.  Maybe use quartz instead?

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
     )
     actual_metadata = get_package_info(inputs[0], metadata=inputs[1])
     assert actual_metadata == expected
+
+
+@pytest.mark.parametrize(
+    "inputs, expected_mu",
+    [
+        # Test whether the function returns the correct mu
+        (  # C1: No density or packing fraction (only for known material), expect to get mu from database
```
sbillinge, line 174: "C1 composition and wavelength provided, known material. Expect to get mu for fully dense material"

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+            {
+                "sample_composition": "H2O",
+                "energy": 10,
+            },
+            0.5330,
+        ),
+        (  # C2: Packing fraction (=0.5) provided only (only for known material)
```
sbillinge, line 181: reuse my pattern from C1 here.

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+            {
+                "sample_composition": "H2O",
+                "energy": 10,
+                "packing_fraction": 0.5,
+            },
+            0.2665,
+        ),
+        (  # C3: Density provided only, expect to compute mu based on it
```
sbillinge, line 189: reuse my pattern above.

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+        ),
+        (  # 2. Unknown material
+            {
+                "sample_composition": "ZrO2",
+                "energy": 17,
+                "density": 1.009,
+            },
+            1.252,
+        ),
+        (  # C4: Both density and packing fraction are provided, expect to compute mu based on both
```
sbillinge, line 206: this doesn't make any sense.  It will result in two numbers, not one number.   We probably want to raise an exception if both are provided.

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+                "density": 1.009,
+                "packing_fraction": 0.5,
+            },
+            0.626,
+        ),
+    ],
+)
+def test_compute_mu_using_xraydb(inputs, expected_mu):
+    actual_mu = compute_mu_using_xraydb(**inputs)
+    assert actual_mu == pytest.approx(expected_mu, rel=0.01, abs=0.1)
```
sbillinge, line 229: this seems a bit too loose?

tests/test_tools.py
```diff
@@ -160,3 +165,65 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+        (  # 2. Unknown material
+            {
+                "sample_composition": "ZrO2",
+                "energy": 17,
+                "density": 1.009,
+                "packing_fraction": 0.5,
+            },
+            0.626,
+        ),
+    ],
```
sbillinge, line 225: we eem to be missing "no density, unknown material".   Here expected could either be to raise an exception or to do a "best effort" calculation with a warning.  I am not sure which is better.  It depends how many unknown materials we have.

sbillinge (commented):
please see inline comments.

sbillinge:
@yucongalicechen further to my review comments, I recommend not using the densities from the xrayDB at all and just computing using given densities.  We can add another function that is `get_density_from_cloud()` or something that tries to get it from materials project or COD.   This would require there to be a packing fraction specified, because the density we got from MP would be the fully dense density, so "material-density" not "sample-density"

yucongalicechen:
> @yucongalicechen further to my review comments, I recommend not using the densities from the xrayDB at all and just computing using given densities. We can add another function that is `get_density_from_cloud()` or something that tries to get it from materials project or COD. This would require there to be a packing fraction specified, because the density we got from MP would be the fully dense density, so "material-density" not "sample-density"

Just to clarify - this means user must enter density or packing fraction right? Something like what this is doing https://11bm.xray.aps.anl.gov/absorb/absorb.php? I can make another issue on getting density.

bobleesj:
https://github.com/diffpy/diffpy.labpdfproc/pull/146#issuecomment-2563799905 - actually I meant to write this comment here in diffpy.utils @yucongalicechen

sbillinge:
> > @yucongalicechen further to my review comments, I recommend not using the densities from the xrayDB at all and just computing using given densities. We can add another function that is `get_density_from_cloud()` or something that tries to get it from materials project or COD. This would require there to be a packing fraction specified, because the density we got from MP would be the fully dense density, so "material-density" not "sample-density"
> 
> Just to clarify - this means user must enter density or packing fraction right? Something like what this is doing https://11bm.xray.aps.anl.gov/absorb/absorb.php? I can make another issue on getting density.

Yes, either or and it triggers a different workflow.

src/diffpy/utils/tools.py
```diff
@@ -206,3 +209,47 @@ def get_package_info(package_names, metadata=None):
         pkg_info.update({package: importlib.metadata.version(package)})
     metadata["package_info"] = pkg_info
     return metadata
+
+
+def compute_mu_using_xraydb(sample_composition, energy, sample_mass_density=None, packing_fraction=None):
```
yucongalicechen, line 214: change name from `density` to `sample_mass_density`, set default of `packing_fraction` to None

src/diffpy/utils/tools.py
```diff
@@ -206,3 +209,47 @@ def get_package_info(package_names, metadata=None):
...
+    """Compute the attenuation coefficient (mu) using the XrayDB database.
+
+    Computes mu based on the sample composition and energy.
+    User should provide a sample mass density or a packing fraction.
+    If neither density nor packing fraction is specified, or if both are specified, a ValueError will be raised.
```
yucongalicechen, line 219: add more description on what to specify

src/diffpy/utils/tools.py
```diff
@@ -206,3 +209,47 @@ def get_package_info(package_names, metadata=None):
...
+    -------
+    mu : float
+        The attenuation coefficient mu in mm^{-1}.
+    """
+    if (sample_mass_density is None and packing_fraction is None) or (
+        sample_mass_density is not None and packing_fraction is not None
+    ):
+        raise ValueError(
+            "You must specify either sample_mass_density or packing_fraction, but not both. "
+            "Please rerun specifying only one."
```
yucongalicechen, line 243: raise an error if user specify none of mass density/packing fraction or both

src/diffpy/utils/tools.py
```diff
@@ -206,3 +209,47 @@ def get_package_info(package_names, metadata=None):
...
+        )
+    if sample_mass_density is not None:
+        mu = material_mu(sample_composition, energy * 1000, density=sample_mass_density, kind="total") / 10
+    else:
+        warnings.warn(
+            "Warning: Density is set to None if a packing fraction is specified, "
+            "which may cause errors for some materials. "
+            "We recommend specifying sample mass density for now. "
+            "Auto-density calculation is coming soon."
```
yucongalicechen, line 252: add a warning message for now. I set density to None if user specifies packing fraction. We can probably replace this with `density=get_density_from_cloud()`. There're some instructions here on how to estimate the density from given chemical formula: https://11bm.xray.aps.anl.gov/absorb/absorb.php. I can try to look into this more.

tests/test_tools.py
yucongalicechen: Most materials are "unknown" I think. I suggest we do a best effort calculation with a warning? I added this as an issue #286.

yucongalicechen:
@sbillinge ready for another review

sbillinge (commented):
please see inline comments.

Let's leave the "materials project/COD" workflow for v3.7.0

to correctly handle that, the best way is to define the functions and their signatures now (so the API desn't change when we implement it) but `raise NotImplmentedError`

Then wrap the line where the function is called wrap it in a try-except clause and handle the `NotImplementedError`, something like:
```
def get_density_from_the_cloud(composition, mp_token="":
    """
    docstring with few words about intent and that it isn't implemented yet
    """"
    raise NotImplementedError
    return
```
then let's say this function gets called in the code, we do this:
```
try:
    material_density = get_density_from_cloud(composition) 
except NotImplementedError:
    warning.warn("some helpful message about retrying using densiy")
```
This is a much tidier way of encapsulating what we need to change when we do actually implement it.

src/diffpy/utils/tools.py
```diff
@@ -206,3 +209,56 @@ def get_package_info(package_names, metadata=None):
         pkg_info.update({package: importlib.metadata.version(package)})
     metadata["package_info"] = pkg_info
     return metadata
+
+
+def get_density_from_cloud(sample_composition, mp_token=""):
+    """Function to get material density from the MP database.
```
yucongalicechen, line 215: add new function - not implemented

src/diffpy/utils/tools.py
```diff
@@ -206,3 +209,56 @@ def get_package_info(package_names, metadata=None):
...
+    ):
+        raise ValueError(
+            "You must specify either sample_mass_density or packing_fraction, but not both. "
+            "Please rerun specifying only one."
+        )
+    if packing_fraction is None:
+        packing_fraction = 1
+        try:
+            sample_mass_density = get_density_from_cloud(sample_composition) * packing_fraction
+        except NotImplementedError:
```
yucongalicechen, line 257: add density function and raise notimplemented error

tests/test_tools.py
yucongalicechen: I didn't add test for implementation error - shall I add one?

bobleesj:
@yucongalicechen for future commit msgs/issues, could we try using this standard?  (also aded to gitlab doc) https://github.com/Billingegroup/cookiecutter/issues/212

src/diffpy/utils/tools.py
sbillinge: Do we want to also go to the COD database (doesn't require a token).

src/diffpy/utils/tools.py
```diff
@@ -242,14 +250,15 @@ def compute_mu_using_xraydb(sample_composition, energy, sample_mass_density=None
...
-            "We recommend specifying sample mass density for now. "
-            "Auto-density calculation is coming soon."
-        )
-        mu = material_mu(sample_composition, energy * 1000, density=None, kind="total") * packing_fraction / 10
+    if packing_fraction is None:
+        packing_fraction = 1
+        try:
+            sample_mass_density = get_density_from_cloud(sample_composition) * packing_fraction
+        except NotImplementedError:
+            warnings.warn(
```
sbillinge, line 258: I think this needs to be an error not a warning, no?  We need to interrupt execution.

Also, let's apologise... "So sorry, Density computation from composition is not implemented right now.   We hope to have this implemented in the next release. Please rerun specifying a sample mass density.

sbillinge (commented):
Nice, thanks!  Please see inline comments.

yucongalicechen:
> @yucongalicechen for future commit msgs/issues, could we try using this standard? (also aded to gitlab doc) [Billingegroup/cookiecutter#212](https://github.com/Billingegroup/cookiecutter/issues/212)

thanks Bob! Will use that in my future commit messages

src/diffpy/utils/tools.py
yucongalicechen: Got it! I will add the COD database and change the error message

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,53 @@ def get_package_info(package_names, metadata=None):
         pkg_info.update({package: importlib.metadata.version(package)})
     metadata["package_info"] = pkg_info
     return metadata
+
+
+def get_density_from_cloud(sample_composition, mp_token=""):
+    """Function to get material density from the MP or COD database.
```
yucongalicechen, line 214: added COD database here

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,53 @@ def get_package_info(package_names, metadata=None):
...
+
+    It is not implemented yet.
+    """
+    raise NotImplementedError(
```
yucongalicechen, line 218: I moved the error msg here

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,53 @@ def get_package_info(package_names, metadata=None):
...
+    """
+    if (sample_mass_density is None and packing_fraction is None) or (
+        sample_mass_density is not None and packing_fraction is not None
+    ):
+        raise ValueError(
+            "You must specify either sample_mass_density or packing_fraction, but not both. "
+            "Please rerun specifying only one."
+        )
+    if packing_fraction is not None:
+        sample_mass_density = get_density_from_cloud(sample_composition) * packing_fraction
```
yucongalicechen, line 257: I think this will be more concise - use packing fraction to compute mass density and pass to `material_mu()`

yucongalicechen:
@sbillinge ready for another review! Thanks so much for the comments above.
I've added the MP/COD workflow in a new issue #286 following the new group standard

tests/test_tools.py
sbillinge: no, we don't need a test for that.

tests/test_tools.py
```diff
@@ -160,3 +165,58 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+def test_compute_mu_using_xraydb(inputs, expected_mu):
+    actual_mu = compute_mu_using_xraydb(**inputs)
+    assert actual_mu == pytest.approx(expected_mu, rel=1e-6, abs=1e-4)
+
+
+@pytest.mark.parametrize(
+    "inputs",
+    [
+        # Test when the function raises ValueError
+        # C1: Both mass density and packing fraction are provided
```
sbillinge, line 198: To compactify we want to write this as follows....also, put what you expect, e.g., 
```
        # Test when the function has invalid inputs
        ( # C1: Both mass density and packing fraction are provided, expect ValueError exception
            {
                "sample_composition": "quartz",
                "energy": 10,
                "sample_mass_density": 2.65,
                "packing_fraction": 1,
            }
        ),
```

tests/test_tools.py
sbillinge: This test makes sense when we implement the getting density from the cloud.

sbillinge (commented):
nice, thanks.  Please see small inline comments

sbillinge (commented):
nearly there.  Still trying to finalize the example for others.

tests/test_tools.py
```diff
@@ -160,3 +165,34 @@ def test_get_package_info(monkeypatch, inputs, expected):
...
+
+
+@pytest.mark.parametrize(
+    "inputs",
+    [
+        # Test when the function has invalid inputs
+        # C1: Both mass density and packing fraction are provided, expect ValueError exception
+        (
+            {
+                "sample_composition": "quartz",
```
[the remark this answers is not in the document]
yucongalicechen: Ah gotcha. Will fix that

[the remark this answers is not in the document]
yucongalicechen: Ooops there're conflicts - willl fix it

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,53 @@ def get_package_info(package_names, metadata=None):
...
+
+    Computes mu based on the sample composition and energy.
+    User should provide a sample mass density or a packing fraction.
+    If neither density nor packing fraction is specified, or if both are specified, a ValueError will be raised.
+    Reference: https://xraypy.github.io/XrayDB/python.html#xraydb.material_mu.
+
+    Parameters
+    ----------
+    sample_composition : str
+        The chemical formula or the name of the material.
```
sbillinge, line 236: not the name, just chemical formula.

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,53 @@ def get_package_info(package_names, metadata=None):
...
+    energy : float
+        The energy of the incident x-rays in keV.
+    sample_mass_density : float, optional, Default is None
+        The mass density of the packed powder/sample in gr/cm^3.
```
sbillinge, line 240: g/cm*3, not gr/cm^3

src/diffpy/utils/tools.py
```diff
@@ -206,3 +208,53 @@ def get_package_info(package_names, metadata=None):
...
+    packing_fraction : float, optional, Default is None
+        The fraction of sample in the capillary (between 0 and 1).
```
sbillinge, line 242: add "specify either `sample_mass_density` or `packing_fraction` but not both."

sbillinge (commented):
please see inline

sbillinge:
nice @yucongalicechen !

PR #280, opened 2024-12-26
improve comments for `test_scale_to`

### LINKED ISSUE #274: Improve comments for @pytest.mark.parametrize test cases in `test_scale_to` function
@sbillinge ready for review.

@yucongalicechen , the part that is remaining is the `test_scale_to` function. 

The existing comments  like`# C2: Same length x-arrays with exact x-value match`, `# C3: Same length x-arrays with approximate x-value match` were abit hard for me to understand what's being tested. Would you mind improving this test comments for `test_scale_to` using the above format?

_Originally posted by @bobleesj in https://github.com/diffpy/diffpy.utils/issues/273#issuecomment-2560272419_

yucongalicechen:
closes #274 
@bobleesj do you wanna give it a read first to see if the comments make sense to you?

### DIFF
--- a/tests/test_diffraction_objects.py
+++ b/tests/test_diffraction_objects.py
@@ -190,7 +190,7 @@ def test_init_invalid_xtype():
 @pytest.mark.parametrize(
     "org_do_args, target_do_args, scale_inputs, expected",
     [
-        # Test whether scale_to() scales to the expected values
+        # Test whether the original y-array is scaled as expected
         (  # C1: Same x-array and y-array with 2.1 offset, expect y-array to shift up by 2.1
             {
                 "xarray": np.array([10, 15, 25, 30, 60, 140]),
--- a/tests/test_diffraction_objects.py
+++ b/tests/test_diffraction_objects.py
@@ -215,7 +215,7 @@ def test_init_invalid_xtype():
         (  # C2: Same x-arrays
             # x-value has exact matches at tth=60 (y=60) and tth=60 (y=6),
             # for original and target diffraction objects,
-            # expect y-array to multiply by 6/60=1/10
+            # expect original y-array to multiply by 6/60=1/10
             {
                 "xarray": np.array([10, 15, 25, 30, 60, 140]),
                 "yarray": np.array([10, 20, 25, 30, 60, 100]),
@@ -239,7 +239,7 @@ def test_init_invalid_xtype():
         (  # C3: Different x-arrays with same length,
             # x-value has closest match at q=0.12 (y=10) and q=0.14 (y=1)
             # for original and target diffraction objects,
-            # expect y-array to multiply by 1/10
+            # expect original y-array to multiply by 1/10
             {
                 "xarray": np.array([0.12, 0.24, 0.31, 0.4]),
                 "yarray": np.array([10, 20, 40, 60]),
@@ -263,7 +263,7 @@ def test_init_invalid_xtype():
         (  # C4: Different x-array lengths
             # x-value has closest matches at tth=61 (y=50) and tth=62 (y=5),
             # for original and target diffraction objects,
-            # expect y-array to multiply by 5/50=1/10
+            # expect original y-array to multiply by 5/50=1/10
             {
                 "xarray": np.array([10, 25, 30.1, 40.2, 61, 120, 140]),
                 "yarray": np.array([10, 20, 30, 40, 50, 60, 100]),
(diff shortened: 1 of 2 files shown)


bobleesj:
> closes #274 
> 
> @bobleesj do you wanna give it a read first to see if the comments make sense to you?

Yup currently shopping right now, I will review in 1-2 hrs

tests/test_diffraction_objects.py
```diff
@@ -191,7 +191,7 @@ def test_init_invalid_xtype():
     "org_do_args, target_do_args, scale_inputs, expected",
     [
         # Test whether scale_to() scales to the expected values
```
bobleesj, line 193: `scales_to() scales to` feels a bit circulating?

tests/test_diffraction_objects.py
```diff
@@ -191,7 +191,7 @@ def test_init_invalid_xtype():
...
-        (  # C1: Same x-array and y-array with 2.1 offset, expect yarray shifted by 2.1 offset
+        (  # C1: Same x-array and y-array with 2.1 offset, expect y-array to shift up by 2.1
```
bobleesj, line 194: clear

tests/test_diffraction_objects.py
```diff
@@ -212,7 +212,7 @@ def test_init_invalid_xtype():
             },
             {"xtype": "tth", "yarray": np.array([4.1, 5.1, 6.1, 7.1, 8.1, 9.1])},
         ),
-        (  # C2: Same length x-arrays with exact x-value match
+        (  # C2: Same x-arrays, x-value matches at tth=60, expect y-array to divide by 10
```
bobleesj, line 215: If i understand correctly, we compare the yarray_1 of 60 against yarray_2 of 6 (located at xarray of 60) and we expect the y-array to e divided by 10 based on this ratio?

Just comment - since both xarray and yarray use 60 for the first input, it's a bit hard to follow

tests/test_diffraction_objects.py
```diff
@@ -233,7 +233,10 @@ def test_init_invalid_xtype():
             },
             {"xtype": "tth", "yarray": np.array([1, 2, 2.5, 3, 6, 10])},
         ),
-        (  # C3: Same length x-arrays with approximate x-value match
+        (  # C3: Different x-arrays with same length,
+            # x-value has closest match at q=0.12 (y=10) and q=0.14 (y=1)
+            # for original and target diffraction objects,
+            # expect y-array to divide by 10
```
bobleesj, line 239: ok - yarray of 10 and 1 are compared since they are the nearest values at q=0.1. Since the second is 1/10 of the original, we expect y value to be 1/10

tests/test_diffraction_objects.py
yucongalicechen: Yes! It finds the closest index in each x-array and scales based on that. In this case the scaling factor is 60/6=10 and the y-array for the original diffraction object is divided by 10

tests/test_diffraction_objects.py
```diff
@@ -254,7 +257,10 @@ def test_init_invalid_xtype():
             },
             {"xtype": "q", "yarray": np.array([1, 2, 4, 6])},
         ),
-        (  # C4: Different x-array lengths with approximate x-value match
+        (  # C4: Different x-array lengths
```
bobleesj, line 260: okay - clear, I understand the function now. yeah for C2, it was hard for me to understand what the function does by reading the test alone since numbers were similiar but other cases are clear.

bobleesj (commented):
@yucongalicechen pls see my comments - thanks a lot Yucong!

tests/test_diffraction_objects.py
yucongalicechen: Thanks! I'll edit C2

tests/test_diffraction_objects.py
yucongalicechen: Oops yes forgot to change this - thanks!

yucongalicechen:
@sbillinge ready for review

sbillinge (commented):
This is great, thanks @yucongalicechen 

My one request might be to reorder the cases so the offset goes to the end.  In general, I think that things would be easiest to understand if we test the main obvious functionality first, then edge cases later.  This way, the reader first gets the idea what the function is supposed to be doing when everything is working first, then starts to dig into funky edge-cases and unusual behavior later when they already have a clear understanding of the main purpose of the function.

@bobleesj maybe we can add this to our "best practices"?   What do you think.  Here the offset is a funky extra functionality that I would rather be at the end.

bobleesj:
> @bobleesj maybe we can add this to our "best practices"? What do you think. Here the offset is a funky extra functionality that I would rather be at the end.

6. Order test cases from the most general to edge cases. This helps readers understand the basic function behavior first before utilizing or encountering unusual features or behaviors.

@sbillinge Yup added to our gitlab page

PR #282, opened 2024-12-27
validators.py created

### LINKED ISSUE #7: factor out isfloat to a separate file
Perhaps to diffpy.utils.typecheck module?

- [ ] check if there is any existing Python package with "isfloat" functionality.

lucas-lemos-da-silva:
close #7 the module validators.py was created with the function isnumber. On the file loaddata.py the old "isfloat" was deleted. Please let me know if something is missing.


bobleesj:
Great @lucas-lemos-da-silva 

We need to have a news file - Could you try to add a news file and i think we have the instruction provided?

We also need to run pre-commit before making a PR, could you also do that, i think we also have that in the instruction.

bobleesj:
Also please refer to the "Lessons Learned" gitlab page on how to design commit msgs and PR titles

sbillinge (commented):
Thanks for this @lucas-lemos-da-silva 

I think we maybe don't want `validtaors.py` inside `parsers` do we?  It is a more boadly useful capability.  

Also, I am not sure that `isnumber()` is really a validator per se.   maybe we want to put it in to tools.py as a tool?  We can discuss.  @bobleesj?

The function seems to be called `isnumber()` but it is actually testing for whether the object can be converted to a float, not a number.

I think we maybe want to start with some tests and then we can work on the function.

I expect that this change in the code API will break things elsewhere too, so I would expect that there are likely other files that need to be updated.   @lucas-lemos-da-silva do tests pass locally after this change?

lucas-lemos-da-silva:
Thank you for the comments! The current `isnumber()` function checks if an object can be converted to a float (almost all numbers). Upon closer inspection, I realized that it excludes complex and imaginary numbers. I am now exploring alternatives. Regarding the local tests, they do not seem to be working. Apologies for that. I am working on identifying the issue.

lucas-lemos-da-silva:
I will close this PR because I did not sync with the `main` before creating a branch. I will continue our discussion in a new PR. @sbillinge I am in touch with @bobleesj right now.

Issue #279, opened 2024-12-26
`get_array_index` returns `list`, instead should be `int`  


bobleesj:
---
let's check.  Logically, this should return an int not a list.  Let's check and maybe update the function/tesnts, or understand why it returns a list and change the name.

_Originally posted by @sbillinge in https://github.com/diffpy/diffpy.utils/pull/271#discussion_r1897870501_

PR #283, opened 2024-12-27
Fix `get_array_index` returns `int` instead of `list` in DiffractionObject

bobleesj:
Closes https://github.com/diffpy/diffpy.utils/issues/279

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -331,29 +331,29 @@ class DiffractionObject:
     def uuid(self, _):
         raise AttributeError(_setter_wmsg("uuid"))
 
-    def get_array_index(self, value, xtype=None):
+    def get_array_index(self, xtype, xvalue):
         """Return the index of the closest value in the array associated with
-        the specified xtype.
+        the specified xtype and the value provided.
 
         Parameters
         ----------
-        xtype str
-            the xtype used to access the array
-        value float
-            the target value to search for
+        xtype : str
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -331,29 +331,29 @@ def uuid(self):
     def uuid(self, _):
         raise AttributeError(_setter_wmsg("uuid"))
 
-    def get_array_index(self, value, xtype=None):
+    def get_array_index(self, xtype, xvalue):
```
bobleesj, line 334: Switched the order for the parameters based on behavior:

user first decide which xtype to search for, and then enter the value second..

src/diffpy/utils/diffraction_objects.py
```diff
@@ -331,29 +331,29 @@ def uuid(self):
...
         """Return the index of the closest value in the array associated with
-        the specified xtype.
+        the specified xtype and the value provided.
```
bobleesj, line 336: small refactoring of variable names and docstring

tests/test_diffraction_objects.py
```diff
@@ -364,7 +364,7 @@ def test_scale_to_bad(org_do_args, target_do_args, scale_inputs):
                 "xtype": "tth",
                 "value": 30.005,
             },
-            [0],
+            0,
```
bobleesj, line 367: int!

bobleesj (commented):
Ready for review @sbillinge

PR #289, opened 2024-12-29
add new features in `scale_to()` so user can run without an x-value

### LINKED ISSUE #287: DiffractionObject.scale_to() default is to scale to maxima
### Problem
Sorry guys, this is a scope creep, but I had an a idea to make the `scale_to()` method more useful so that it can be run with no xvalue passed and still work by default. 

### Proposed solution
 It would be something like this:
```
scaled_do = my_do.scaled_to(target_do)
```
and the default scaling would simply be `max(target_do.on_q()[0])/max(my_do.on_q()[0])`

if user passes in an xvalue, and xtype we go to the current workflow.  This allows the user to override the default behavior when it doesn't work for some reason, for example they want to scale to a diffraction feature but there is small angle scattering in the signal, or the maximum peak in one pattern is a different peak or something.

yucongalicechen:
closes #287
@sbillinge reafy for review

### DIFF
--- a/doc/source/examples/diffraction_objects_example.rst
+++ b/doc/source/examples/diffraction_objects_example.rst
@@ -117,6 +117,13 @@ For convenience, you can also apply an offset to the scaled new diffraction obje
 
     scaled_and_offset_measured = measured.scale_to(calculated, q=5.5, offset=0.5)
 
+You can call `scale_to()` without specifying a value for `q`, `tth`, or `d`.
+In this case, the scaling will be done based on the maximal x-array value of both diffraction objects:
+
+.. code-block:: python
+
+    scaled_measured = measured.scale_to(calculated)
+
 DiffractionObject convenience functions
 ---------------------------------------
 
--- a/doc/source/examples/diffraction_objects_example.rst
+++ b/doc/source/examples/diffraction_objects_example.rst
@@ -106,6 +106,16 @@ we would replace the code above with
 The ``scale_to()`` method returns a new ``DiffractionObject`` which we can assign to a new
 variable and make use of,
 
+The default behavior is to align the objects based on the maximal q-value of each diffraction object,
+so they will align at the intensity at these indices.
+
+.. code-block:: python
+
+    scaled_measured = measured.scale_to(calculated)
+
+If this doesn't give the desirable results, you can specify an ``xtype=value`` to scale
+based on the closest x-value in both objects. For example:
+
 .. code-block:: python
 
     scaled_measured = measured.scale_to(calculated, q=5.5)
@@ -116,13 +126,6 @@ For convenience, you can also apply an offset to the scaled new diffraction obje
 .. code-block:: python
 
     scaled_and_offset_measured = measured.scale_to(calculated, q=5.5, offset=0.5)
-
-You can call `scale_to()` without specifying a value for `q`, `tth`, or `d`.
-In this case, the scaling will be done based on the maximal x-array value of both diffraction objects:
-
-.. code-block:: python
-
-    scaled_measured = measured.scale_to(calculated)
 
 DiffractionObject convenience functions
 ---------------------------------------
--- a/news/scaleto-max.rst
+++ b/news/scaleto-max.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* new feature in `scale_to()` to run without specifying an x-value
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/scaleto-max.rst
+++ b/news/scaleto-max.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* new feature in `scale_to()` to run without specifying an x-value
+* new feature in `scale_to()`: default scaling is based on the max q-value in each diffraction object.
 
 **Changed:**
 
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -408,22 +408,22 @@ class DiffractionObject:
         Parameters
         ----------
         target_diff_object: DiffractionObject
-            the diffraction object you want to scale the current one onto
+            The diffraction object you want to scale the current one onto.
 
-        q, tth, d : float, optional, default is the max q-value from each object
+        q, tth, d : float, optional, default is None
             The value of the x-array where you want the curves to line up vertically.
             Specify a value on one of the allowed grids, q, tth, or d), e.g., q=10.
 
         offset : float, optional, default is None
-            an offset to add to the scaled y-values
+            The offset to add to the scaled y-values.
 
         Returns
         -------
-        the rescaled DiffractionObject as a new object
+        scaled : DiffractionObject
+            The rescaled DiffractionObject as a new object.
         """
         if offset is None:
             offset = 0
-
         scaled = self.copy()
         count = sum([q is not None, tth is not None, d is not None])
         if count > 1:
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -400,9 +400,8 @@ class DiffractionObject:
         """Returns a new diffraction object which is the current object but
         rescaled in y to the target.
 
-        By default, if none of `q`, `tth`, or `d` are provided,
-        the scaling is based on the maximal x-array value from both objects.
-        The y-value in the target at the closest specified x-value will be used as the factor to scale to.
+        By default, if `q`, `tth`, or `d` are not provided, scaling is based on the max q-value from each object.
+        Otherwise, y-value in the target at the closest specified x-value will be used as the factor to scale to.
         The entire array is scaled by this factor so that one object places on top of the other at that point.
         If multiple values of `q`, `tth`, or `d` are provided, an error will be raised.
 
@@ -411,7 +410,7 @@ class DiffractionObject:
         target_diff_object: DiffractionObject
             the diffraction object you want to scale the current one onto
 
-        q, tth, d : float, optional, default is q with the maximal x-array value from each object
+        q, tth, d : float, optional, default is the max q-value from each object
             The value of the x-array where you want the curves to line up vertically.
             Specify a value on one of the allowed grids, q, tth, or d), e.g., q=10.
 
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -433,7 +433,9 @@ class DiffractionObject:
             )
 
         if count == 0:
-            scaled._all_arrays[:, 0] *= max(target_diff_object.on_q()[1]) / max(self.on_q()[1])
+            q_target_max = max(target_diff_object.on_q()[1])
+            q_self_max = max(self.on_q()[1])
+            scaled._all_arrays[:, 0] = scaled._all_arrays[:, 0] * q_target_max / q_self_max + offset
             return scaled
 
         xtype = "q" if q is not None else "tth" if tth is not None else "d"
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -400,7 +400,7 @@ class DiffractionObject:
         """Returns a new diffraction object which is the current object but
         rescaled in y to the target.
 
-        By default, if `q`, `tth`, or `d` are not provided, scaling is based on the max q-value from each object.
+        By default, if `q`, `tth`, or `d` are not provided, scaling is based on the max intensity from each object.
         Otherwise, y-value in the target at the closest specified x-value will be used as the factor to scale to.
         The entire array is scaled by this factor so that one object places on top of the other at that point.
         If multiple values of `q`, `tth`, or `d` are provided, an error will be raised.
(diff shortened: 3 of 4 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -402,14 +402,16 @@ def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=0):
 
         The y-value in the target at the closest specified x-value will be used as the factor to scale to.
         The entire array is scaled by this factor so that one object places on top of the other at that point.
-        If multiple values of `q`, `tth`, or `d` are provided, or none are provided, an error will be raised.
+        If none of `q`, `tth`, or `d` are provided,
+        the scaling will be based on the maximal x-array value from both objects.
```
yucongalicechen, line 406: add new feature

src/diffpy/utils/diffraction_objects.py
```diff
@@ -402,14 +402,16 @@ def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=0):
...
+        If multiple values of `q`, `tth`, or `d` are provided, an error will be raised.
 
         Parameters
         ----------
         target_diff_object: DiffractionObject
             the diffraction object you want to scale the current one onto
 
-        q, tth, d : float, optional, must specify exactly one of them
+        q, tth, d : float, optional, default is q with the maximal x-array value of the current object
```
yucongalicechen, line 414: edit description

src/diffpy/utils/diffraction_objects.py
```diff
@@ -422,11 +424,16 @@ def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=0):
         """
         scaled = self.copy()
         count = sum([q is not None, tth is not None, d is not None])
-        if count != 1:
+        if count > 1:
             raise ValueError(
-                "You must specify exactly one of 'q', 'tth', or 'd'. Please rerun specifying only one."
+                "You must specify none or exactly one of 'q', 'tth', or 'd'. "
```
yucongalicechen, line 429: edit error msg

src/diffpy/utils/diffraction_objects.py
```diff
@@ -422,11 +424,16 @@ def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=0):
...
+                "Please provide either none or one value."
             )
 
+        if count == 0:
+            scaled._all_arrays[:, 0] *= max(target_diff_object.on_q()[1]) / max(self.on_q()[1])
+            return scaled
```
yucongalicechen, line 435: when no xvalue was specified

tests/test_diffraction_objects.py
```diff
@@ -284,23 +284,7 @@ def test_init_invalid_xtype():
...
-    # Check the intensity data is the same as expected
-    assert np.allclose(scaled_do.on_xtype(expected["xtype"])[1], expected["yarray"])
-
-
-@pytest.mark.parametrize(
-    "org_do_args, target_do_args, scale_inputs",
-    [
-        # Test expected errors produced from scale_to() with invalid inputs
-        (  # C1: none of q, tth, d, provided, expect ValueError
+        (  # C5: none of q, tth, d, provided, expect to scale on the maximal x-arrays
```
yucongalicechen, line 287: new test case

doc/source/examples/diffraction_objects_example.rst
```diff
@@ -117,6 +117,13 @@ For convenience, you can also apply an offset to the scaled new diffraction obje
 
     scaled_and_offset_measured = measured.scale_to(calculated, q=5.5, offset=0.5)
 
+You can call `scale_to()` without specifying a value for `q`, `tth`, or `d`.
```
sbillinge, line 120: This will be the default behavior so should go first in the docs.    Then we can say something like "if this doesnlt give desirable results you can user....." and describe the other workflow.

To give even more flexibility to the user we may want to allow them to specify a separate xtype and xvalue for the target and the scaled DO.  The hierarchy woiuld be:
1. no xtype/xvalue
2. if that doesn't give desirable results, specify xtype and xvalue for target.  scale to the same (nearest) point in the scaled pattern
3. if that doesn't give desirable results, specify xtype, and a different xvalue for target. and scaled DO, respectively.  Scale the yvale of xvalue_scaled to yvalue of xvalue_target.

This last functionality is a new one and we should put it on a new issue (and a new PR if we want to implement it)

news/scaleto-max.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* new feature in `scale_to()` to run without specifying an x-value
```
sbillinge, line 3: describe what it does, not what it doesn't do...i.e., scales to max value in each chase.  You can also add that it allows this to be used without xvalue

src/diffpy/utils/diffraction_objects.py
sbillinge: as above.  Put this behavior first.

tests/test_diffraction_objects.py
sbillinge: put this one first?  I think it is worth the extra effort as this is our new default behavior.

tests/test_diffraction_objects.py
```diff
@@ -319,8 +303,25 @@ def test_scale_to(org_do_args, target_do_args, scale_inputs, expected):
                 "d": None,
                 "offset": 0,
```
sbillinge, line 304: this is a different issue, unrelated to what you are doing here, but I would rather have default offset to be None (set to zero if None).  I think this is more logical for my brain.  If I don't set an offset, the offset is None.

tests/test_diffraction_objects.py
```diff
@@ -319,8 +303,25 @@ def test_scale_to(org_do_args, target_do_args, scale_inputs, expected):
...
+            {"xtype": "q", "yarray": np.array([10, 20, 30])},
         ),
-        (  # C2: tth and d both provided, expect ValueErrort
+    ],
+)
+def test_scale_to(org_do_args, target_do_args, scale_inputs, expected):
+    original_do = DiffractionObject(**org_do_args)
+    target_do = DiffractionObject(**target_do_args)
+    scaled_do = original_do.scale_to(
+        target_do, q=scale_inputs["q"], tth=scale_inputs["tth"], d=scale_inputs["d"], offset=scale_inputs["offset"]
```
sbillinge, line 314: I think we need to fgure out a way only pass in what is needed, rather than passing in everything with some things set to default values.  It may need more test functions written, but see if you can figure out how to do it with unpacking kwargs that are handed in.

sbillinge (commented):
Thanks @yucongalicechen !. please see inline comments.

yucongalicechen:
@sbillinge ready for review

doc/source/examples/diffraction_objects_example.rst
```diff
@@ -106,6 +106,16 @@ we would replace the code above with
 The ``scale_to()`` method returns a new ``DiffractionObject`` which we can assign to a new
 variable and make use of,
 
+The default behavior is to align the objects based on the maximal q-value of each diffraction object,
+so they will align at the intensity at these indices.
+
+.. code-block:: python
+
+    scaled_measured = measured.scale_to(calculated)
+
```
yucongalicechen, line 115: put default behavior first

news/scaleto-max.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* new feature in `scale_to()`: default scaling is based on the max q-value in each diffraction object.
```
yucongalicechen, line 3: edited

src/diffpy/utils/diffraction_objects.py
```diff
@@ -396,37 +396,46 @@ def on_tth(self):
     def on_d(self):
         return [self.all_arrays[:, 3], self.all_arrays[:, 0]]
 
-    def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=0):
+    def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=None):
```
yucongalicechen, line 399: change default offset to None

src/diffpy/utils/diffraction_objects.py
```diff
@@ -396,37 +396,46 @@ def on_tth(self):
...
         """Returns a new diffraction object which is the current object but
         rescaled in y to the target.
 
-        The y-value in the target at the closest specified x-value will be used as the factor to scale to.
+        By default, if `q`, `tth`, or `d` are not provided, scaling is based on the max q-value from each object.
+        Otherwise, y-value in the target at the closest specified x-value will be used as the factor to scale to.
```
yucongalicechen, line 404: put default behavior first in docstring

src/diffpy/utils/diffraction_objects.py
yucongalicechen: If user specified an offset do we want to add offset to the scaled array?

tests/test_diffraction_objects.py
```diff
@@ -276,22 +277,15 @@ def test_init_invalid_xtype():
...
         ),
     ],
 )
 def test_scale_to(org_do_args, target_do_args, scale_inputs, expected):
     original_do = DiffractionObject(**org_do_args)
     target_do = DiffractionObject(**target_do_args)
-    scaled_do = original_do.scale_to(
-        target_do, q=scale_inputs["q"], tth=scale_inputs["tth"], d=scale_inputs["d"], offset=scale_inputs["offset"]
-    )
+    scaled_do = original_do.scale_to(target_do, **scale_inputs)
```
yucongalicechen, line 288: edited - this allows us to only pass in what we need

src/diffpy/utils/diffraction_objects.py
```diff
@@ -396,32 +396,34 @@ def on_tth(self):
...
-        the scaling will be based on the maximal x-array value from both objects.
         If multiple values of `q`, `tth`, or `d` are provided, an error will be raised.
 
         Parameters
         ----------
         target_diff_object: DiffractionObject
             the diffraction object you want to scale the current one onto
 
-        q, tth, d : float, optional, default is q with the maximal x-array value of the current object
+        q, tth, d : float, optional, default is the max q-value from each object
```
sbillinge, line 413: "default is None"

src/diffpy/utils/diffraction_objects.py
```diff
@@ -396,32 +396,34 @@ def on_tth(self):
...
             The value of the x-array where you want the curves to line up vertically.
             Specify a value on one of the allowed grids, q, tth, or d), e.g., q=10.
 
-        offset : float, optional, default is 0
+        offset : float, optional, default is None
             an offset to add to the scaled y-values
```
sbillinge, line 418: "The offset to be added to the scaled y-values"

src/diffpy/utils/diffraction_objects.py
```diff
@@ -396,32 +396,34 @@ def on_tth(self):
...
 
         Returns
         -------
         the rescaled DiffractionObject as a new object
         """
+        if offset is None:
+            offset = 0
+
```
sbillinge, line 426: remove empty line

sbillinge (commented):
small tweaks

src/diffpy/utils/diffraction_objects.py
sbillinge: yes, that is correct.  Add the offset to the y-array after scaling

tests/test_diffraction_objects.py
```diff
@@ -255,15 +261,10 @@ def test_init_invalid_xtype():
...
-                "offset": 0,
-            },
+            {"tth": 60},
             {"xtype": "tth", "yarray": np.array([1, 2, 3, 4, 5, 6, 10])},
         ),
-        (  # C4: Same x-array and y-array with 2.1 offset, expect y-array to shift up by 2.1
+        (  # C5: Same x-array and y-array with 2.1 offset, expect y-array to shift up by 2.1
             {
                 "xarray": np.array([10, 15, 25, 30, 60, 140]),
                 "yarray": np.array([2, 3, 4, 5, 6, 7]),
```
sbillinge, line 270: This test is compromised because the target and scale y-arrays are the same, so it is and edge case.  In general, we want to scale, then offset.

sbillinge (commented):
please see inline.  I answered a question, and raised a new one.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -396,37 +396,48 @@ def on_tth(self):
     def on_d(self):
         return [self.all_arrays[:, 3], self.all_arrays[:, 0]]
 
-    def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=0):
+    def scale_to(self, target_diff_object, q=None, tth=None, d=None, offset=None):
         """Returns a new diffraction object which is the current object but
         rescaled in y to the target.
 
-        The y-value in the target at the closest specified x-value will be used as the factor to scale to.
+        By default, if `q`, `tth`, or `d` are not provided, scaling is based on the max intensity from each object.
```
yucongalicechen, line 403: change max q-value to max intensity here

src/diffpy/utils/diffraction_objects.py
yucongalicechen: added the offset here

tests/test_diffraction_objects.py
```diff
@@ -255,43 +261,48 @@ def test_init_invalid_xtype():
...
-                "tth": 60,
-                "d": None,
-                "offset": 0,
-            },
+            {"tth": 60},
             {"xtype": "tth", "yarray": np.array([1, 2, 3, 4, 5, 6, 10])},
         ),
-        (  # C4: Same x-array and y-array with 2.1 offset, expect y-array to shift up by 2.1
+        (  # C5.1: Reuse test case from C1, none of q, tth, d, provided, but include an offset,
+            # expect scaled y-array in C1 to shift up by 2
```
yucongalicechen, line 268: add tests for offset, reuse patterns from C1 and C4 so that they're more general

yucongalicechen:
@sbillinge ready for review!

Issue #291, opened 2024-12-29
docs: add docs for mu / muD calculator


yucongalicechen:
add docs for #281, #286, #219, #45

sbillinge:
User docs are needed.   Target audience is a chemist who wants to use the code to get a PDF from their lab data using PDFgetX3.  These folks will never be looking at API docs, nor even using an API.

PR #292, opened 2024-12-29
feat: muD calculator

### LINKED ISSUE #281: muD calculator for z-scan files
move mud_calculator.py from labpdfproc to here

Check https://github.com/diffpy/diffpy.labpdfproc/blob/main/src/diffpy/labpdfproc/mud_calculator.py
and https://github.com/diffpy/diffpy.labpdfproc/blob/main/tests/test_mud_calculator.py.

yucongalicechen:
closes #281 
This is a direct copy and paste from labpdfproc
I added a new issue for docs in #291
@sbillinge ready for review

### DIFF
--- a/src/diffpy/utils/tools.py
+++ b/src/diffpy/utils/tools.py
@@ -3,25 +3,11 @@ import json
 from copy import copy
 from pathlib import Path
 
+import numpy as np
+from scipy.optimize import dual_annealing
+from scipy.signal import convolve
 
-def clean_dict(obj):
-    """Remove keys from the dictionary where the corresponding value is None.
-
-    Parameters
-    ----------
-    obj: dict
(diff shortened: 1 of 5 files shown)


src/diffpy/utils/tools.py
```diff
@@ -3,25 +3,11 @@
 from copy import copy
 from pathlib import Path
 
+import numpy as np
+from scipy.optimize import dual_annealing
+from scipy.signal import convolve
 
-def clean_dict(obj):
```
yucongalicechen, line 7: looks like this function is not used anywhere else.. so I deleted it

bobleesj:
Screenshot 2024-12-29 at 6 18 05 PM

@yucongalicechen nice!!

PR #303, opened 2024-12-31
fix: remove user_config.py

yucongalicechen:
I don't think this is used anywhere, so I deleted it
@sbillinge ready for review


sbillinge (commented):
which functions are doing these tasks?

news/remove-userconfig.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* `user_config.py` function.
```
bobleesj, line 15: Since this is written for changelog.rst, both used by savvy users and future maintainers (could be new members) maybe a bit more context would be better as to what was changed (user config.py) which is stated and also maybe why this was  removed?

yucongalicechen:
> which functions are doing these tasks?

`_load_config` and `check_and_build_global_config`.

yucongalicechen:
> > which functions are doing these tasks?
> 
> `_load_config` and `check_and_build_global_config` in tools.

yucongalicechen:
@sbillinge ready for review. I clarified in the news.

sbillinge:
actually, I think we should keep this and use it.  In principle, we can allow users to override the name and path of the config file this way.

sbillinge:
I took a look at this, and I think it doesn't add much.  I will merge this and we can get rid of it.  The other functions can be easily extended if we want to allow users to override the paths or filenames.

PR #307, opened 2024-12-31
docs: add docstrings for on_d, on_q, on_tth, and dump

### LINKED ISSUE #306: docs: docstring for `on_d`, `on_q`, `on_tth`, and `dump` are missing
@sbillinge Do we want to make `on_d`, `on_q`, `on_tth` private?

yucongalicechen:
closes #306 
@sbillinge ready for review

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -418,12 +418,33 @@ class DiffractionObject:
             return self.on_d(), "d"
 
     def on_q(self):
+        """Return the tuple of two 1D numpy arrays containing q and y data.
+
+        Returns
+        -------
+        (xarray, yarray) : tuple of ndarray
+            The tuple containing two 1D numpy arrays with q and y data
+        """
         return [self.all_arrays[:, 1], self.all_arrays[:, 0]]
 
     def on_tth(self):
+        """Return the tuple of two 1D numpy arrays containing tth and y data.
+
+        Returns
+        -------
+        (xarray, yarray) : tuple of ndarray
+            The tuple containing two 1D numpy arrays with tth and y data
+        """
         return [self.all_arrays[:, 2], self.all_arrays[:, 0]]
 
     def on_d(self):
+        """Return the tuple of two 1D numpy arrays containing d and y data.
+
+        Returns
+        -------
+        (xarray, yarray) : tuple of ndarray
+            The tuple containing two 1D numpy arrays with d and y data
+        """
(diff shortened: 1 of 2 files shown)


src/diffpy/utils/diffraction_objects.py
```diff
@@ -418,12 +418,33 @@ def _get_original_array(self):
             return self.on_d(), "d"
 
     def on_q(self):
+        """Return the tuple of two 1D numpy arrays containing q and y data.
+
+        Returns
+        -------
+        (xarray, yarray) : tuple of ndarray
```
sbillinge, line 425: let's change `xarray` to `q-array` to be explicit.  And also below, xarray-> `tth-array` and so on.  Just to make it really clear.

src/diffpy/utils/diffraction_objects.py
```diff
@@ -507,6 +528,16 @@ def on_xtype(self, xtype):
             raise ValueError(_xtype_wmsg(xtype))
 
     def dump(self, filepath, xtype=None):
+        """Dump the xarray and yarray of the diffraction object to a two-column
+        file, with the associated information included in the header.
+
+        Parameters
+        ----------
+        filepath : str
```
sbillinge, line 536: non programmers may not know that that the path includes the filename.  Best here is to give an example.

sbillinge (commented):
brilliatn @yucongalicechen .  Please see a couple of small edits.

src/diffpy/utils/diffraction_objects.py
yucongalicechen: @sbillinge added in the new commit. please review

src/diffpy/utils/diffraction_objects.py
```diff
@@ -537,6 +537,14 @@ def dump(self, filepath, xtype=None):
             The filepath where the diffraction object will be dumped
         xtype : str, optional, default is q
             The type of quantity for the independent variable chosen from {*XQUANTITIES, }
+
+        Examples
+        --------
+        To save a diffraction object to a file named "diffraction_data.chi" with
+        the independent variable 'q':
+
+        >>> file = "diffraction_data.chi"
```
sbillinge, line 546: Let's also have another example, that shows a relative path.  I think that should then be clear enough.

sbillinge (commented):
nice!  One last request.

src/diffpy/utils/diffraction_objects.py
yucongalicechen: Sounds great! I'll add that.

PR #326, opened 2025-02-03
update flake8, black and isort to set line length limit to 79

ycexiao:
Closes #324


ycexiao:
@sbillinge It's ready for review

sbillinge:
Thanks, that was a big one on the end but we'll with doing

PR #333, opened 2025-04-18
feat: implement `get_density_from_cloud()`

### LINKED ISSUE #331: feat: `get_density_from_cloud()` COD workflow

yucongalicechen:
closes #331 
@sbillinge ready for review

### DIFF
--- a/src/diffpy/utils/tools.py
+++ b/src/diffpy/utils/tools.py
@@ -4,6 +4,7 @@ from copy import copy
 from pathlib import Path
 
 import numpy as np
+from pymatgen.ext.cod import COD
 from scipy.optimize import dual_annealing
 from scipy.signal import convolve
 from xraydb import material_mu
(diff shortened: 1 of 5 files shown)


src/diffpy/utils/tools.py
```diff
@@ -214,21 +215,48 @@ def get_package_info(package_names, metadata=None):
...
-        "right now. "
-        "We hope to have this implemented in the next release. "
-        "Please rerun specifying a sample mass density."
+    results = COD().get_structure_by_formula(sample_composition)
+    if phase:
+        phase = phase.replace(" ", "").lower()
+        for r in results:
+            if r.get("sg", "unknown").replace(" ", "").lower() == phase:
+                return r["structure"].density
+    print(
```
yucongalicechen, line 239: Print out available densities and phases. Since COD() handles invalid samples and phases I didn't write any error message here.

sbillinge (commented):
That's nicely done but it is failing tests.  We may want to mock the API call in the test.  We generally don't want the tests to rely on an internet connection.

yucongalicechen:
> That's nicely done but it is failing tests. We may want to mock the API call in the test. We generally don't want the tests to rely on an internet connection.

I got a chance to play around with the functions a bit more. I also meant to say "space group" instead of "phase" earlier and in the code. I think the issues are:
1. pymatgen removed some densities by default according to this: https://github.com/materialsproject/pymatgen/pull/3419, so it's missing a few common ones like SiO2. I'm wondering if it is better to call MP API directly, just so we have some fallback/default values?
2. It looks like there can be multiple densities for the same sample composition and space group. Not sure how to pick a default in that case? I think the order is a bit random and that's why the test failed here. We could also give users the option to specify space group & ID?

sbillinge:
honestly I am not sure I would rely on pymatgen/MP.   Why not just make an API call to COD and get the cif and compute the density?

PR #335, opened 2025-05-02
feat: add function that converts chemical formula to hill notation

yucongalicechen:
Task 1 in #333 
@sbillinge ready for review! I splitted the issue into multiple steps, and this is the first one. For this issue I think this might be a better way to organize the substeps since everything is in one place.

### DIFF
--- a/news/hill-notation.rst
+++ b/news/hill-notation.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Function that converts a chemical formula to hill notation.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/src/diffpy/utils/tools.py
+++ b/src/diffpy/utils/tools.py
@@ -1,5 +1,7 @@
 import importlib.metadata
 import json
+import re
+from collections import defaultdict
 from copy import copy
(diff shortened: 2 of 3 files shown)


sbillinge (commented):
I am not sure this is the right approach.  I think it might be more reusable if we take a `diffpy.structure` object and write a method that will output hill notation.

yucongalicechen:
> I am not sure this is the right approach. I think it might be more reusable if we take a `diffpy.structure` object and write a method that will output hill notation.

I think this function is only for converting the chemical formula to hill notation though. After the user specifies the chemical formula, I'm thinking of the following steps: 1. convert to hill notation, 2. use it to extract relevant cif files, 3. compute density for each cif file, 4. return the average density (or other depending on phase). I plan to have one function each for each step. Does this sound good?

sbillinge:
> > I am not sure this is the right approach. I think it might be more reusable if we take a `diffpy.structure` object and write a method that will output hill notation.
> 
> I think this function is only for converting the chemical formula to hill notation though. After the user specifies the chemical formula, I'm thinking of the following steps: 1. convert to hill notation, 2. use it to extract relevant cif files, 3. compute density for each cif file, 4. return the average density (or other depending on phase). I plan to have one function each for each step. Does this sound good?

I was just suggesting that this function could be a method of `diffpy.structure`.  This makes sense if we can more or less instantiate an empty diffpy.structure object.  A diffpy.structure object is a structure or a phase, so it makes sense that I can ask it for its hill_formula......

PR #336, opened 2025-05-02
feat: add function that returns cif filenames based on chemical formula

yucongalicechen:
Task 2 in #331
@sbillinge ready for review

### DIFF
--- a/news/get-cif.rst
+++ b/news/get-cif.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Function that returns a list of cif filenames based on a given chemical formula.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/requirements/conda.txt
+++ b/requirements/conda.txt
@@ -1,4 +1,3 @@
 numpy
 xraydb
 scipy
-bs4
--- a/requirements/conda.txt
+++ b/requirements/conda.txt
@@ -1,3 +1,4 @@
 numpy
 xraydb
 scipy
+bs4
--- a/requirements/pip.txt
+++ b/requirements/pip.txt
@@ -1,4 +1,3 @@
 numpy
 xraydb
 scipy
-bs4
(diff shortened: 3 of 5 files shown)


sbillinge (commented):
is there a reason you are using beautiful soup and not making an API call to COD?

sbillinge (commented):
that looks better.  What is the payload from the API call?  I am a bit confused that there are cif files involved.  Doesn't it send json or sthg like that?

yucongalicechen:
> that looks better. What is the payload from the API call? I am a bit confused that there are cif files involved. Doesn't it send json or sthg like that?

Yes, the API returns a json file. I implemented something similar to what I saw here https://github.com/materialsproject/pymatgen/blob/master/src/pymatgen/ext/cod.py. I was thinking it might be cleaner if we extract the cif filenames, so that when we compute the density we can pass the cif filenames as arguments. If we keep the functions for computing density private, it would be easier to work with the json files directly?

sbillinge (commented):
this looks good.  Two quick questions:
1) why are we dealing with cif files?   Is there a way to get the data without files involved?
2) Let's think about whether we want to use the pymatgen wrapper directly.  Downside is we then have pymatgen as a dependency.  Another Downside is that we can then only do things that pymatgetn can do.  Upside is that they maintain that code.  Let's just think about it for a bit.

But issue (1) is more problematic for me if we can avoid it.

yucongalicechen:
> this looks good. Two quick questions:
> 
> 1. why are we dealing with cif files?   Is there a way to get the data without files involved?
> 2. Let's think about whether we want to use the pymatgen wrapper directly.  Downside is we then have pymatgen as a dependency.  Another Downside is that we can then only do things that pymatgetn can do.  Upside is that they maintain that code.  Let's just think about it for a bit.
> 
> But issue (1) is more problematic for me if we can avoid it.

We don't need the cif files, from the purpose of computing density we can stop at ``data = response.json()`` and use the json file directly. At this point I don't think we need ``pymatgen`` because it removed quite a lot of densities for common chemical compositions.

yucongalicechen:
> > We don't need the cif files, from the purpose of computing density we can stop at `data = response.json()` and use the json file directly. At this point I don't think we need `pymatgen` because it removed quite a lot of densities for common chemical compositions.
> 
> Shall we close this then, since we don't need this function?

Yep, closing this, made a new PR on computing density based on a json file in #338

PR #338, opened 2025-05-13
feat: Compute density for a given cif file

yucongalicechen:
Address Task 3 in #331.
@sbillinge ready for review


src/diffpy/utils/tools.py
```diff
@@ -214,6 +217,49 @@ def get_package_info(package_names, metadata=None):
     return metadata
 
 
+def compute_density_from_cif(sample_composition, cif_data):
+    """Compute the theoretical density from given CIF metadata.
```
yucongalicechen, line 221: This function computes the theoretical density from a single cif json file retrived from COD API. I haven't taken care of the errors yet (for example, missing data from json file). Not sure what is the best approach here but I'm thinking about returning N/A for density when any error happens?

src/diffpy/utils/tools.py
```diff
@@ -214,6 +217,49 @@ def get_package_info(package_names, metadata=None):
     return metadata
 
 
+def compute_density_from_cif(sample_composition, cif_data):
```
sbillinge, line 220: maybe `material_density_from_cif`?

src/diffpy/utils/tools.py
```diff
@@ -214,6 +217,49 @@ def get_package_info(package_names, metadata=None):
...
+    """Compute the theoretical density from given CIF metadata.
+
+    Parameters
+    ----------
+    sample_composition : str
+        The chemical formula of the material, e.g. "NaCl".
+    cif_data : dict
```
sbillinge, line 227: We should think about this.  I wonder if it makes sense that we carry material data around always in structure object.  We could use `diffpy.structure`, but if we are moving in that direction we could use ASE structure containers.

Then we would like to pass around always whatever our "standard" structure object is and we get rid of mention of cif everywhere.   We then would have loaders that loaded cif from file into our structure container (SC), and that loaded COD-json into SC, and pymatgent whatever int SC and so on.  Since this function is, given a structure, calculate a density it should be taking our base structure object and returning density (it doesn't need to take chemical formula obviously since this is in the SC

src/diffpy/utils/tools.py
```diff
@@ -214,6 +217,49 @@ def get_package_info(package_names, metadata=None):
     return metadata
 
 
+def compute_density_from_cif(sample_composition, cif_data):
+    """Compute the theoretical density from given CIF metadata.
+
+    Parameters
+    ----------
+    sample_composition : str
```
sbillinge, line 225: maybe `chemical_formula` for the variable name?

sbillinge (commented):
please see comments.  This is not quite right yet.

PR #340, opened 2025-06-10
skpkg: getting package up to `scikit-package` CI standards

cadenmyers13:
Most files are default files modified by skpkg. no pre-commit edits were necessary.

### DIFF
--- a/src/diffpy/utils/__init__.py
+++ b/src/diffpy/utils/__init__.py
@@ -12,7 +12,7 @@
 # See LICENSE.rst for license information.
 #
 ##############################################################################
-"""General utilities for analyzing diffraction data"""
+"""General utilities for analyzing diffraction data."""
 
 # package version
 from diffpy.utils.version import __version__  # noqa


pyproject.toml
```diff
@@ -6,28 +6,28 @@ build-backend = "setuptools.build_meta"
...
+keywords = ['text data parsers', 'wx grid', 'diffraction objects']
 readme = "README.rst"
 requires-python = ">=3.11, <3.14"
 classifiers = [
         'Development Status :: 5 - Production/Stable',
         'Environment :: Console',
         'Intended Audience :: Developers',
         'Intended Audience :: Science/Research',
-        'License :: Free To Use But Restricted',
+        'License :: OSI Approved :: BSD License',
```
sbillinge, line 23: please check the license to make sure it is BSD before accepting this change.

src/diffpy/__init__.py
```diff
@@ -16,14 +12,3 @@
 # See LICENSE.rst for license information.
 #
 ##############################################################################
-"""diffpy - tools for structure analysis by diffraction.
```
sbillinge, line 19: its unlikely that we want to delete the contents of this file.  But I am not sure.  Is this a new development that we can handle the namespace strcture (diffpy.something) without this extend path?  We may want to ask @bobleesj

tests/conftest.py
```diff
@@ -1,112 +1,19 @@
...
-        "owner_orcid": "home_orcid",
-    }
+
+    home_config_data = {"username": "home_username", "email": "home@email.com"}
     with open(home_dir / "diffpyconfig.json", "w") as f:
         json.dump(home_config_data, f)
-    yield home_dir, cwd_dir
-
-
-@pytest.fixture
```
sbillinge, line 27: do you want to delete all our tests?

tests/test_version.py
```diff
@@ -1,6 +1,6 @@
 """Unit tests for __version__.py."""
 
-import diffpy.utils
+import diffpy.utils  # noqa
```
sbillinge, line 3: It is probably ok to leave these `# noqa`'s but they are not needed.  I think we need them in the template because there is jinja code there that black can;t handle but after the template is built it would be better if they were deleted (but not hte end of theworld if they stay.

sbillinge (commented):
pls see comments

pyproject.toml
cadenmyers13: @sbillinge confirmed BSD from skpkg

sbillinge (commented):
please see comment

tests/test_version.py
sbillinge: no shell script, but maybe something in the `post-gen-hook`

cadenmyers13: @bobleesj Do you have any insight on how this might be done?

src/diffpy/__init__.py
bobleesj: An empty `src/<namespacename>/__init__.py` is still needed.

The namespace's``__init__.py`, it's written in `post_gen_hook.py` shown below: 

```
def __gen_init__(module_name):
    """Generate __init__.py file for namespace module."""
    __init__ = f"""#!/usr/bin/env python
##############################################################################
#
# (c) {% now 'utc', '%Y' %} The Trustees of Columbia University in the City of New York.
# All rights reserved.
#
# File coded by: Billinge Group members and community contributors.
#
# See GitHub contributions for a more detailed list of contributors.
# https://github.com/{{ cookiecutter.github_username_or_orgname }}/{{ cookiecutter.github_repo_name }}/graphs/contributors
#
# See LICENSE.rst for license information.
#
##############################################################################

""" # noqa: E999
    return __init__
```

tests/test_version.py
bobleesj: Why did you add ` # noqa`? The cookiecutter template doesn't require it:

```
"""Unit tests for __version__.py."""

import {{ cookiecutter.package_dir_name }}  # noqa


def test_package_version():
    """Ensure the package version is defined and not set to the initial
    placeholder."""
    assert hasattr({{ cookiecutter.package_dir_name }}, "__version__")
    assert {{ cookiecutter.package_dir_name }}.__version__ != "0.0.0"
``

bobleesj (commented):
@sbillinge @cadenmyers13 replied to your comments

sbillinge:
regarding the empty init, this was our question.  Is just an empty init needed.  Before in diffpy.utils we had some code about extending paths or something.  I vaguely remember we managed to delete that and the namespace stuff all still worked, but I just wanted confirmation that my memory was correct, we don't need the delete "extend path" logic.

bobleesj:
> The # noqa is left over from hte template.(line 3)

Ok. I see it. 

>  Before in diffpy.utils we had some code about extending paths or something

Yes, we deleted it and it works without it - both via pytest after deployed to pypi/conda-forge.

tests/test_version.py
cadenmyers13: @bobleesj That was already there so it generated with `package create public`

PR #356, opened 2025-07-25
Fix: let `DiffractionObject.get_array_index` to use `xtype` from its inputs.

### LINKED ISSUE #355: fix: `get_array_index` ignores the input `xtype`
### Problem

The function `DiffractionObject.get_array_index` source code is
```
    def get_array_index(self, xtype, xvalue):
        """Return the index of the closest value in the array associated with
        the specified xtype and the value provided.
        ...
        """
        xtype = self._input_xtype
        xarray = self.on_xtype(xtype)[0]
        if len(xarray) == 0:
            raise ValueError(
                f"The '{xtype}' array is empty. "
                "Please ensure it is initialized."
            )
        index = (np.abs(xarray - xvalue)).argmin()
        return index
```

So the input parameter `xtype` doesn't have an effect at all. Thus, the following code
```
x = np.linspace(0, 90, 1000)
y = np.random.rand(1000)
tmp = DiffractionObject(
     xarray=x,
     yarray=y,
     xtype="angle",
     wavelength=1.54,
)
index_angle = tmp.get_array_index(xvalue=0.24, xtype="angle")
index_q_from_function = tmp.get_array_index(xvalue=0.24, xtype="q")
real_index_q =  np.argmin(np.abs(tmp.on_xtype("q")[0] - 0.24))
print(index_angle == index_q_from_function)
print(index_q_from_function == real_index_q)
```
returns True and False.

ycexiao:
### What problem does this PR address

Closes #355 

### What should the reviewer(s) do

Please check the tests and implementation

### DIFF
--- a/src/diffpy/utils/diffraction_objects.py
+++ b/src/diffpy/utils/diffraction_objects.py
@@ -419,7 +419,7 @@ class DiffractionObject:
             The value of the xtype to find the closest index for.
         xtype : str, optional
             The type of the independent variable in `xarray`. Must be one
-            of {*XQUANTITIES,}. Default is {self._input_xtype}
+            of {*XQUANTITIES, }. Default is {self._input_xtype}
 
         Returns
         -------


tests/test_diffraction_objects.py
```diff
@@ -376,8 +376,8 @@ def test_scale_to_bad(org_do_args, target_do_args, scale_inputs):
             },
             0,
         ),
-        (  # C2: Target value lies in the array, expect the (first) closest
-            # index
+        # C2: Target value lies in the array, expect the closest index
+        (  # 1. same xtype
```
sbillinge, line 380: what does "same xtype" mean? same as what?

sbillinge (commented):
I feel that the unfinished discussion is the behavior that we want and the function signature.  The way the docstring is written it should be `get_array_index(value, xtype=None)`, i.e., value is required but xtype not.  `If xtype is None: xtype=_initial_xtype`.

Arguably, we could make xtype required.  I think that this may make sense.  I am not sure the UC where it is implicit.   What do you think?

tests/test_diffraction_objects.py
ycexiao: It means that the `xtype` of the original DiffractionObject and the `xtype` of the data from which the `get_array_index` gets the array index are the same.

I will modify the comment.

sbillinge: This may not need to be changed depending on the outcome of the discussion above.  I am thinking that we may want to make both variables required.  Do you know of a particular UC that requires the other?  Do we now anywhere where this is being used actually?

ycexiao:
> Arguably, we could make xtype required. I think that this may make sense. I am not sure the UC where it is implicit. What do you think?

I think `xtype` should be optional. `DiffractionObject` keeps a property `_input_xtype`.  The expected behavior is that when `xtype` is not specified, `diffpy.utils` will operate on the data in its original format.

sbillinge:
On reflection, let's leave it as it is.  I don't think there is a strong reason to change it and I may have had a reason I wanted it that way.

sbillinge:
Next, for the tests, can we dynamically load a list of xtypes that will automatically update in the future as we add more?  Is there a way to do that?  It would be good if that list could also be dynamically loaded in the docstring.

tests/test_diffraction_objects.py
ycexiao: > Do you know of a particular UC that requires the other?

No. I think what the function does is to find the index. `xtype` is not actually needed. 

And if users want to locate a specific `q` or `tth` that is different from the `_input_xtype`, they can use `diff_obj.on_xtype("<xtype>")`. Also, this might not be a common need.

> Do we now anywhere where this is being used actually?

At least in `diffpy.utils`, it is not referenced. It is just a helper function we provide to users.

ycexiao (commented):
@sbillinge, It's ready for review.

sbillinge:
Thanks for catching that and fixing the tests to catch the bad case

PR #365, opened 2025-11-26
fix: add real data into example

stevenhua0320:
@sbillinge Ready to review


cadenmyers13:
@sbillinge I sent @stevenhua0320 a .chi file and a .fq file too. I think the .chi file is what we want.

cadenmyers13:
@stevenhua0320 Its been a while since I looked at this but I think the gist is to add the data to this repo so that users can grab it and run it for this example. try to fix the example and make it easy for users to run (add a link to the data or something like that).

PR #370, opened 2025-12-21
fix: Remove API for calculating mu from packing fraction

### LINKED ISSUE #339: fix: Remove ``pakcing_fraction`` option from ``compute_mu_from_xraydb``
### Problem

The function is doing too much at this point. We will only accept sample mass density moving forward.

cadenmyers13:
closes #339 

I dont believe a test is needed since this is basically a wrapper for for `material_mu` in `xraydb`.

### DIFF
--- a/news/rm-packing-fraction.rst
+++ b/news/rm-packing-fraction.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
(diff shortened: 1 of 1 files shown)


sbillinge:
nice!

PR #377, opened 2026-02-03
V3.7.0

cadenmyers13:
closes #376

### DIFF
--- a/news/dep-msg-helper.rst
+++ b/news/dep-msg-helper.rst
@@ -1,6 +1,8 @@
 **Added:**
 
 * Add ``build_deprecation_message`` helper for printing consistent deprecation messages.
+* Add ``generate_deprecation_docstring`` for generating a template docstring for deprecated functions.
+
 
 **Changed:**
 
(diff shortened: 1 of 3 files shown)


src/diffpy/utils/_deprecator.py
```diff
@@ -131,40 +131,3 @@ def build_deprecation_message(
         f"version {removal_version}. Please use '{new_base}.{new_name}' "
         f"instead."
     )
-
-
-def generate_deprecation_docstring(new_name, removal_version, new_base=None):
```
sbillinge: why did you remove this?

sbillinge (commented):
please see comments

src/diffpy/utils/_deprecator.py
cadenmyers13: It isnt used anywhere. my idea with it was to have it as a helper that could easily print the docstring then the user could copy/paste the docstring where it needs to go. I removed it because I felt like this was overkill but if you think its helpful we can add it back

PR #384, opened 2026-03-17
Re-add Python 3.10 support

cadenmyers13:


cadenmyers13:
@sbillinge ready for review

PR #386, opened 2026-05-29
chore: updated pre-commit hook dependencies

SYMSCAE:
## Summary

This PR performs pre-commit hook maintenance by updating tooling hook revisions.

## Changes

- Updated `black` from `26.3.1` to `26.5.1`
- Updated `isort` from `8.0.1` to `9.0.0a3`


.pre-commit-config.yaml
```diff
@@ -64,3 +64,4 @@ repos:
       - id: docformatter
         additional_dependencies: [tomli]
         args: [--in-place, --config, ./pyproject.toml]
+        language_version: python3.13
```
sbillinge: What is this? We don't want any pins like this is possible. If a pin is necessary we need an issue to remove it in the future. But better to find a fix that is not a workaround

stevenhua0320: We made this is because now we decide to use `1.7.7` docformatter but this version does not support python `3.14`, so we add the pin to the python version here. We had to wait for the `1.7.8` wierd deleting line bug to fix so that we could remove the pin and I think everything would work smoothly again. So, we should make an issue and notify future developers to remove this pin and change to `1.7.8` version when the bug is fixed. I also made the update to the issue #385

sbillinge: ok, we can't use 1.7.7 then.

sbillinge: we will have to find another workaround with 1.7.8

